PR #3, opened 2024-05-21
TS: Change enums to more idiomatic typescript

jhugman:

This PR proposes a change to how we represent enums with values in Typescript.

[Rendered documentation](https://github.com/jhugman/uniffi-bindgen-react-native/blob/f50fcce5a0a2c36fb89206a67281cac83bb8f841/docs/differences/enums.md).

### DIFF
--- a/crates/uniffi-bindgen-react-native/src/bindings/react_native/gen_typescript/templates/EnumTemplate.ts
+++ b/crates/uniffi-bindgen-react-native/src/bindings/react_native/gen_typescript/templates/EnumTemplate.ts
@@ -42,89 +42,31 @@ const {{ ffi_converter_name }} = (() => {
 })();
 
 {% else %}
-/// Typescript doesn't have the concept of enums with variant properties,
-/// so we need to be a little creative here.
-///
-/// For a Rust enum:
-/// ```rs
-/// enum FilePath {
-///   Local { path: String }
-///   Remote { host: String, path: String }
-/// }
-/// ```
-/// Currently:
-/// ```ts
-/// class FilePath {
-///   static Local = class Local extends FilePath { constructor(public members: { path: string })} {}
-///   static Remote = class Remote extends FilePath { constructor(public members: { host: string, path: string })} {}
-/// }
-/// ```
-///
-/// This gives nice construction properties:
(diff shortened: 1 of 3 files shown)


crates/uniffi-bindgen-react-native/src/bindings/react_native/gen_typescript/templates/EnumTemplate.ts
```diff
@@ -42,104 +42,44 @@ const {{ ffi_converter_name }} = (() => {
...
-            {%- endif %}
-        }
-    }
-    {%- endfor %}
-}
+export type {{ type_name }} = {# space #}
+{%- for variant in e.variants() %}
+{%-   call ts::docstring(variant, 4) %}
+    { kind: {{ kind_type_name }}.{{ variant|variant_name }}
+    {%- if !variant.fields().is_empty() %}, value: {
```
zzorba, line 60: I believe that technically in typescript, you separate type definitions with `;` instead of `,`.  It accepts both because 'javascript' but I'm pretty sure this should be `; value:`

crates/uniffi-bindgen-react-native/src/bindings/react_native/gen_typescript/templates/EnumTemplate.ts
```diff
@@ -42,104 +42,44 @@ const {{ ffi_converter_name }} = (() => {
...
+        {%- for field in variant.fields() %}
+        {%- call ts::field_name(field, loop.index) %}: {{ field|type_name(ci) }}
+        {%- if !loop.last %}, {% endif -%}
```
zzorba, line 63: Semi here too instead of comma

zzorba (approved):
Looks great.

One small style-nit on Typescript syntax, which I'm sure prettier would have eventually complained about.

crates/uniffi-bindgen-react-native/src/bindings/react_native/gen_typescript/templates/EnumTemplate.ts
jhugman: wow. TIL.

PR #4, opened 2024-05-22
Generate compilable typescript for a complicated crate

jhugman:
According to [The Big O of Code Reviews](https://www.egorand.dev/the-big-o-of-code-reviews/), this is a O(_m*n_) change.

Generate enough typescript to compile the FFI crate of the matrix-rust-sdk.

This includes: 

- External types.
- Promises
- Callback Interfaces
- Timestamp, Duration and ByteBuffer types

The most interesting of these commits are likely: 

- 5e1a424 which arranges the order of the generated typescript.
- 8a191f4 which collects "external" types and their converters imported from other "namespaces".


crates/uniffi-bindgen-react-native/src/bindings/react_native/gen_typescript/filters.rs
```diff
@@ -117,6 +117,7 @@ pub fn ffi_type_name_for_cpp(type_: &FfiType, is_internal: &bool) -> Result<Stri
     })
 }
 
+#[allow(unused)]
```
jhugman, line 120: These will go once we re-enable the CallbackInterface and/or promises code.

crates/uniffi-bindgen-react-native/src/bindings/react_native/gen_cpp/filters.rs
```diff
@@ -15,25 +16,24 @@ pub fn ffi_type_name_to_rust(ffi_type: &FfiType) -> Result<String, askama::Error
...
+        FfiType::Float64 => "double".into(),
+        FfiType::RustArcPtr(_) => "void *".into(),
+        FfiType::RustBuffer(_) => "RustBuffer".into(),
+        FfiType::ForeignBytes => "ForeignBytes".into(),
+        FfiType::Callback(_) => "Callback".into(),
+        FfiType::Struct(nm) => format!("Uniffi{}", nm.to_upper_camel_case()), // XXX
+        FfiType::Handle => "Handle".into(),
         FfiType::RustCallStatus => todo!(),
-        FfiType::Reference(_) => todo!(),
+        FfiType::Reference(inner) => format!("{}* _Nonnull", ffi_type_name(inner)?), // XXX
```
jhugman, line 36: These two lines marked `XXX` are the real change here. They're marked with XXX because this is C++ and hasn't been run– I don't think they'd compile yet.

crates/uniffi-bindgen-react-native/src/bindings/react_native/gen_cpp/filters.rs
```diff
@@ -15,25 +16,24 @@ pub fn ffi_type_name_to_rust(ffi_type: &FfiType) -> Result<String, askama::Error
...
+        FfiType::Int32 => "int32_t".into(),
+        FfiType::UInt64 => "uint64_t".into(),
+        FfiType::Int64 => "int64_t".into(),
+        FfiType::Float32 => "float".into(),
+        FfiType::Float64 => "double".into(),
+        FfiType::RustArcPtr(_) => "void *".into(),
+        FfiType::RustBuffer(_) => "RustBuffer".into(),
+        FfiType::ForeignBytes => "ForeignBytes".into(),
+        FfiType::Callback(_) => "Callback".into(),
+        FfiType::Struct(nm) => format!("Uniffi{}", nm.to_upper_camel_case()), // XXX
```
zzorba, line 33: Just for my own knowledge, what is the // XXX here, just a marker to revisit?

crates/uniffi-bindgen-react-native/src/bindings/react_native/gen_typescript/oracle.rs
```diff
@@ -147,7 +148,7 @@ impl CodeOracle {
             FfiType::Callback(name) => self.ffi_callback_name(name),
             FfiType::Struct(name) => self.ffi_struct_name(name),
             FfiType::Reference(inner) => self.ffi_type_label_by_reference(inner),
-            FfiType::VoidPointer => "biginh".to_string(),
+            FfiType::VoidPointer => "/*pointer*/ bigint".to_string(),
```
zzorba, line 151: placeholder?  I think long-term wrapping pointers in a class for TS/JS might be the right call.

crates/uniffi-bindgen-react-native/src/bindings/react_native/gen_typescript/templates/CallbackInterfaceImpl.ts
```diff
@@ -1,68 +1,72 @@
...
+            const makeCall = {# space #}
+            {%- if meth.is_async() %}async {% else %} {% endif-%}
+            (): {% call ts::return_type(meth) %} => {
+                // Was converter.handleMap.get(uniffiHandle);
+                const uniffiObj = {{ ffi_converter_name }}.lift(uniffiHandle);
+                if (uniffiObj === undefined) {
+                    throw new UniffiInternalError.UnexpectedStaleHandle()
                 }
-                return {% if meth.throws() %}try {% endif %}{% if meth.is_async() %}await {% endif %}uniffiObj.{{ meth.name()|fn_name }}(
+                return {% if meth.is_async() %}await {% endif %}uniffiObj.{{ meth.name()|fn_name }}(
```
zzorba, line 28: so the await here is fine and totally acceptable, but not actually required if you are returning a Promise call directly.

I'm pretty sure Javascript/Typescript lets you return either Promise<Foo> or Foo in an async and will handle unwrapping the Promise.

crates/uniffi-bindgen-react-native/src/bindings/react_native/gen_typescript/templates/CallbackInterfaceImpl.ts
```diff
@@ -1,68 +1,72 @@
...
+                    {{ arg|lift_fn }}({{ arg.name()|var_name }}){% if !loop.last %},{% endif %}
                     {%- endfor %}
                 )
             }
             {%- if !meth.is_async() %}
 
             {% match meth.return_type() %}
             {%- when Some(t) %}
-            let writeReturn = { uniffiOutReturn.pointee = {{ t|lower_fn }}($0) }
+            const writeReturn = (obj: any) => { uniffiOutReturn.pointee = {{ t|lower_fn }}(obj) };
```
zzorba, line 38: Is any a permanent thing here, or a temporary workaround -- not meant as a leading question, I understand that any-s are sometimes needed when building low-level bridges (and other systems are giving you typesafety here, so its not even really that bad).

zzorba (approved):
Couple small questions, but was able to follow most stuff (the async stuff confused me a bit, but it also appears to still be mostly a stub function, so I'm not worried about that).

crates/uniffi-bindgen-react-native/src/bindings/react_native/gen_typescript/oracle.rs
jhugman: Yes, I have that in my head as a task called: **Make the JS garbage collector to tell Rust when an object goes out of scope**.

Note to self: start filing GH issues.

crates/uniffi-bindgen-react-native/src/bindings/react_native/gen_cpp/filters.rs
jhugman: Yes, precisely so. My comment [below](https://github.com/jhugman/uniffi-bindgen-react-native/pull/4/files/4925eb51c8c791bed7cd9de39aa7c38dd093671e#r1610368990) may or may not be useful.

crates/uniffi-bindgen-react-native/src/bindings/react_native/gen_typescript/templates/CallbackInterfaceImpl.ts
jhugman: No, this who file is commented out. I'll definitely be revisiting this file.

PR #5, opened 2024-05-23
Refactor and add --ts-out and --cpp-out to command line

jhugman:
According to [The Big O of Code Reviews](https://www.egorand.dev/the-big-o-of-code-reviews/), this is a O(_m*n_) change. (with a small _m_)

This PR works on the command line of the `uniffi-bindgen-react-native`.

It changes it to use a subcommand for the existing functionality:

Previously:

```sh
uniffi-bindgen-react-native --out-dir ./generated --library target/debug/libfoo.dylib
```

Now:

```sh
uniffi-bindgen-react-native bindings --ts-out ./src/generated --cpp-out ./cpp/generated --library target/debug/libfoo.dylib
```

As part of this work, it turned out to be running the typescript formatter on the same files multiple times; this resulted in a refactor of the formatting, and rationalizing some duplicated code (from `xtask` and `uniffi-bindgen-react-native` crates) into the `uniffi_common` crate.

### DIFF
--- a/crates/uniffi-bindgen-react-native/src/bindings/mod.rs
+++ b/crates/uniffi-bindgen-react-native/src/bindings/mod.rs
@@ -3,4 +3,106 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/
  */
+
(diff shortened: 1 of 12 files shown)


zzorba (approved):
LGTM

PR #6, opened 2024-05-24
Export FfiConverters as default exports

jhugman:

This PR makes the change from FfiConverters being exposed as `uniffiConverters` in each of the generated modules to being exported as the default.

i.e.

```ts
import { FooBar } from './foo';

import uniffiConverters from './foo';
const { FfiConerterTypeFooBar } = uniffiConverters;
```

In the `index.ts` of the crates binding, this allows us to do:

```ts
export { * } from './foo'
export { * } from './bar'
```

so as to export all the generated API, but without the converters– and implementation detail of uniffi.


zzorba (approved):
LGTM

PR #8, opened 2024-05-24
Proposal: Typescript for Records with defaults

jhugman:

This PR adds a proposed change to the way we specify records in typescript.

[Rendered documentation](https://github.com/jhugman/uniffi-bindgen-react-native/blob/055fb6d3950621d4165fd094a776a398d3df0503/docs/differences/records.md).


zzorba (approved):
I like this compromise a lot, feels very typescript and anyone familiar with react will feel very comfortable with this idea of named arguments + defaults.

PR #9, opened 2024-05-28
Hooking up GC to drop

jhugman:

This PR reworks the API for objects:

- It renames the method `destroy()` to `uniffiDestroy()`.
- It removes/hides the method `uniffiClonePointer()`.

It also adds a coverall testing fixture, which will become more exercised over the next little while.

The intent of the PR was to create a `RustArcPtr` object which contains a pointer and a destructor callback. [This commit][ts] implements it in Typescript,  followed by [this one in C++][cpp].

[ts]: https://github.com/jhugman/uniffi-bindgen-react-native/commit/70611ee
[cpp]: https://github.com/jhugman/uniffi-bindgen-react-native/commit/afd1697f3de92baff9b5351dd9d581541b49b24f

The utility of this is two fold: 
1. move all JS access to the "raw" pointer to one place where we can check that it (and thus the Rust object) still exists before calling into the Rust, and 
1. call the destructor callback when the JS object is garbage collected.

The first piece works just fine, and the rest of the PR is worth landing. However, [the second piece][cpp] does not.

This is why:

```cpp
  // The C++ object's dtor will be called when the GC finalizes this
  // object.  (This may be as late as when the Runtime is shut down.)
  // You have no control over which thread it is called on.  This will
  // be called from inside the GC, so it is unsafe to do any VM
  // operations which require a Runtime&.  Derived classes' dtors
  // should also avoid doing anything expensive.  Calling the dtor on
  // a jsi object is explicitly ok.  If you want to do JS operations,
  // or any nontrivial work, you should add it to a work queue, and
  // manage it externally.
  virtual ~HostObject();
```

Alternatives available to explore: instead of going via the JS back into C++ and to the Rust, then miss out the JS. Unfortunately, this means the C++ generator knowing how to tie together a `bless` `FfiFunction` with a `free` `FfiFunction`, which at present it does not.

I'm leaving [the C++ commit][cpp] in for review purposes, and then proposing we drop it before merging.


cpp/includes/RustArcPtr.h
```diff
@@ -0,0 +1,90 @@
...
+      return uniffi_jsi::Bridging<uint64_t>::toJs(rt, pointer);
+    } else if (propName == "d") {
+      return jsi::Value(rt, *destructor);
+    }
+    return jsi::Value::undefined();
+  }
+};
+
+template <> struct Bridging<void *> {
+  static jsi::Value bless_pointer(jsi::Runtime &rt, const jsi::Value &pointer_v,
```
zzorba, line 66: Just for my own knowledge, what does the term Bless mean in regard to pointers in this context?

crates/uniffi-bindgen-react-native/src/bindings/react_native/gen_typescript/templates/ObjectTemplate.ts
```diff
@@ -84,11 +83,35 @@ export class {{ impl_class_name }} implements {{ protocol_name }}, UniffiObjectI
...
-}
+const {{ obj_factory }}: UniffiObjectFactory<{{type_name}}> = {
+    create(pointer: UnsafeMutableRawPointer): {{ type_name }} {
+        const instance = Object.create({{ type_name }}.prototype);
+        instance._rustArcPtr = this.bless(pointer);
+        return instance;
+    },
+
+    bless(pointer: UnsafeMutableRawPointer): UniffiRustArcPtr {
+        const destructor = this.freePointer;
```
zzorba, line 94: Not totally sure, but this looks like it might need to be 'bound' to the this pointer to function correct, if its going to be called as a standalone callback function:

`bind(this, this.freePointer)`

Unless the `this` is going to be injected as the first argument when it is called by `uniffiBlessPointer`, otherwise the this object will be missing.

zzorba: I might be misunderstanding, it just has the smell of something that needs a `bind`

crates/uniffi-bindgen-react-native/src/bindings/react_native/gen_typescript/templates/ObjectTemplate.ts
```diff
@@ -84,11 +83,35 @@ export class {{ impl_class_name }} implements {{ protocol_name }}, UniffiObjectI
...
+        }
+        return ptr.p;
+    },
+
+    clonePointer(obj: {{ type_name }}): UnsafeMutableRawPointer {
+        const pointer = this.pointer(obj);
+        return rustCall(callStatus => NativeModule.{{ obj.ffi_object_clone().name() }}(pointer, callStatus) );
+    },
+
+    freePointer(pointer: UnsafeMutableRawPointer): void {
```
zzorba, line 111: Actually, looking at freePointer it doesn't seem to actually use any `this` methods, so it might actually not need the bind after all?  Though I don't know if javascript will have trouble with the missing hidden-first-argument for `this` that it expects.

zzorba (approved):
One question about nomenclature and one *possible* issue around the currying of how I saw the destructor being passed.

Also curious if there's a path where the Hermes GC will drive this, or if that github issue makes it seem like not-the-way to go?

cpp/includes/RustArcPtr.h
jhugman: I couldn't think of a good word for what I was doing: taking a bigint/pointer, and turning it into a _special_ object that acted in a way unlike any other type of object within JS.

I landed on `bless`, which is [Perl's keyword](https://perldoc.perl.org/functions/bless) to make a value into an object with methods.

crates/uniffi-bindgen-react-native/src/bindings/react_native/gen_typescript/templates/ObjectTemplate.ts
jhugman: You're right: the destructor doesn't need the `this`. In this case, `this` when the `freePointer` function is called from C++ is unset. AFAIK, javascript has it as the `this` value, rather than a hidden-first-argument.

jhugman:
> Also curious if there's a path where the Hermes GC will drive this, or if that github issue makes it seem like not-the-way to go?

I think where we got to for Android is that [we should hook up the GC machinery to Rust](https://github.com/mozilla/uniffi-rs/pull/1869), but leave the escape hatch in so that if resources need to be dropped explicitly, they can be.

cpp/includes/RustArcPtr.h
zzorba: Makes sense, just was curious about the history of the name

PR #11, opened 2024-05-29
Typescript was unhappy when publishing the matrix-rust-sdk

zzorba:
The complexity of the anonymous classes was presenting challenges for Typescript compiler when trying to publish the matrix-rust-sdk.

The fix was to generate local classes, and refer to them in the static class, like so:
```
export class AuthenticationException extends Error {
  constructor(message: string) {
    super(message);
  }
  static ClientMissing: typeof _AuthenticationException_ClientMissing;

  static InvalidServerName: typeof _AuthenticationException_InvalidServerName;
   ...
}
class _AuthenticationException_ClientMissing extends AuthenticationException {
  constructor(message: string) {
    super(message);
  }
}
AuthenticationException.ClientMissing = _AuthenticationException_ClientMissing;

class _AuthenticationException_InvalidServerName extends AuthenticationException {
  constructor(message: string) {
    super(message);
  }
}
AuthenticationException.InvalidServerName =
  _AuthenticationException_InvalidServerName;
```

I don't know if the use of underscores is best here, but it seemed like something was needed to avoid name collisions at the top level.  Open to other conventions, since these aren't exported I figured it didn't matter *that* much?

Also found two errors with imports of Arc pointer types, where it wanted it to be `import type`.  It seems like the generated ones were already correct, there were just two hard-coded ones.

### DIFF
--- a/crates/uniffi-bindgen-react-native/src/bindings/react_native/gen_typescript/templates/ErrorTemplate.ts
+++ b/crates/uniffi-bindgen-react-native/src/bindings/react_native/gen_typescript/templates/ErrorTemplate.ts
@@ -7,6 +7,7 @@ export class {{ type_name }} extends Error {
     }
     {%- if e.is_flat() %}
     {%-   for variant in e.variants() %}
+    {%-    call ts::docstring(variant, 4) %}
     {%-    let var_name = variant.name()|class_name(ci) %}
     static {{ var_name }}: typeof _{{ type_name }}_{{ var_name }};
     {% endfor -%}
@@ -16,7 +17,6 @@ export class {{ type_name }} extends Error {
 }
 {%- if e.is_flat() %}
 {%-   for variant in e.variants() %}
-{%-    call ts::docstring(variant, 4) %}
 {%-    let var_name = variant.name()|class_name(ci) %}
 class _{{ type_name }}_{{ var_name }} extends {{ type_name }} {
     constructor(message: string) { super(message); }
--- a/crates/uniffi-bindgen-react-native/src/bindings/react_native/gen_typescript/templates/ErrorTemplate.ts
+++ b/crates/uniffi-bindgen-react-native/src/bindings/react_native/gen_typescript/templates/ErrorTemplate.ts
@@ -2,21 +2,29 @@
 
 {%- call ts::docstring(e, 0) %}
 export class {{ type_name }} extends Error {
-    private constructor(message: string) {
+    constructor(message: string) {
         super(message);
     }
     {%- if e.is_flat() %}
     {%-   for variant in e.variants() %}
-    {%-    call ts::docstring(variant, 4) %}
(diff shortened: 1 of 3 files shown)


crates/uniffi-bindgen-react-native/src/bindings/react_native/gen_typescript/templates/ErrorTemplate.ts
```diff
@@ -2,21 +2,29 @@
...
     {%- endif %}
 }
+{%- if e.is_flat() %}
+{%-   for variant in e.variants() %}
+{%-    call ts::docstring(variant, 4) %}
+{%-    let var_name = variant.name()|class_name(ci) %}
+class _{{ type_name }}_{{ var_name }} extends {{ type_name }} {
+    constructor(message: string) { super(message); }
+}
+{{ type_name }}.{{ var_name }} = _{{ type_name }}_{{ var_name }};
```
jhugman, line 24: I bow to your superior typescript-fu: couple of questions:

1. does the docstring travel with the class definition to the static variable declaration? i.e. if type `OuterError.InnerVariant` and hover over `InnerVariant` in my editor, does it show the doc string?
2. would putting all these class definitions in an IIFE confuse the whatever was getting confused more, or would it be a valid way of not cluttering the namespace to avoid having the `_{{ type_name }}_` and just use the `{{ var_name }}`?

crates/uniffi-bindgen-react-native/src/bindings/react_native/gen_typescript/templates/ErrorTemplate.ts
```diff
@@ -2,21 +2,29 @@
...
+{% endfor -%}
```
jhugman, line 25: Nit: for complicated templates, I've found adding the indentation after the `{%`; so

```
{%- if condition %}
{%-   for v in variants %}
{%-     let name = v.name()|class_name %}
{%-     call do_stuff(v) %}
…
{%-   endfor %}
{%- endif %}
```

i.e. 
```suggestion
{%-  endfor %}
```

jhugman (approved):
Looks good! A couple of questions, but nothing to hold you back from landing.

crates/uniffi-bindgen-react-native/src/bindings/react_native/gen_typescript/templates/ErrorTemplate.ts
zzorba: I think you are correct, it should probably be on the static declarations. I will move that down.

As for question 2, it appeared that the Typescript compiler was unhappy about the number of anonymous inner classes.  The definitions that has a small number of errors worked, but those that had 10+ variants were exploding due to 'complexity'.  It seemed like by moving the class definition outside of the class (so they were no longer inner classes but proper global ones), it stopped being an issue.

PR #12, opened 2024-06-04
Build and export JS and *.d.ts files

zzorba:
Attempt to try to export *.js and *.d.ts files so library can be consumed.

It doesn't appear to fix my issue unfortunately, but still might be a good thing to do?  It's possible Copilot was misleading me here.

I think if we do this we do need to add a build step somehow.


jhugman (approved):
Does this require a build step now? Could this be added to `package.json`, e.g. `yarn dist` or `yarn publish`.

zzorba:
Added a prepare step, which I believe will run before publish and after install from git directly (if copilot can be trusted)

jhugman:
LGTM!

PR #14, opened 2024-06-06
Make C++ buildable from iOS

jhugman:

This is a small addition to add the *.h files to Xcode for depending libraries.

Until we add this to a pod spec repository, the apps using depending libraries will need to add the following line to their Podfiles:

```ruby
pod 'uniffi-bindgen-react-native', :path => './node_modules/uniffi-bindgen-react-native'
```

After we add this to a pod spec repository, the depending libraries can add:

```ruby
s.dependency 'uniffi-bindgen-react-native'
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,10 @@
   "homepage": "https://github.com/jhugman/uniffi-bindgen-react-native",
   "repository": "https://github.com/jhugman/uniffi-bindgen-react-native.git",
   "license": "MPL-2.0",
-  "author": "James Hugman <james@hugman.tv>",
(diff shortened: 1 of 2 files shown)


zzorba:
LGTM.  

The relative import in your first example might run into issues in mono-repos, where it might hoist dependencies.

I think this code would work to find the actual location of uniffi-bindgen-react-native, which could then be fed to your podspec line above:
```File.dirname(`node --print "require.resolve('uniffi-bindgen-react-native/package.json')"`)```

zzorba (approved):

PR #17, opened 2024-06-11
Mechanical refactor: prefixing directories and crates with ubrn

jhugman:

This is a mechanical refactor to change the names of the crates to `ubrn_common`, `ubrn_cli` and `ubrn_bindgen`.

This is for multiple reasons:
- it rationalizes all related crates to a single prefix
- the `cli` can now be called `uniffi-bindgen-react-native` to outside callers
- it separates out the crates coming from `uniffi-rs`.

Concretely, this PR consists of three steps:
- Rename directories to prefix with ubrn
- Rename uniffi-common to ubrn_common
- Rename cli command to uniffi-bindgen-react-native

Every other change is a consequence of these three steps.


zzorba:
LGTM

PR #18, opened 2024-06-13
Use callback to get the ffi NativeModule

jhugman:


zzorba (approved):
LGTM

crates/ubrn_bindgen/src/bindings/react_native/gen_typescript/templates/wrapper-without-codegen.ts
```diff
@@ -12,8 +12,7 @@ interface NativeModuleInterface {
 
 // Casting globalThis to any allows us to look for `{{ config.cpp_module }}`
 // if it was added via JSI.
-//
-// The empty object is there for testing purposes only, and may be removed.
-export default ((globalThis as any).{{ config.cpp_module}} ?? {}) as NativeModuleInterface;
+const getter: () => NativeModuleInterface = () => (globalThis as any).{{ config.cpp_module }};
```
zzorba, line 15: not required, but a comment here explaining why its a getter might help some poor unfortunate soul someday who decides to optimize it away.

PR #20, opened 2024-06-18
Command line to generate other files 

jhugman:
According to [The Big O of Code Reviews](https://www.egorand.dev/the-big-o-of-code-reviews/), this is a O(_m*n_) change.

This has several changes:

- we change the way Android libs are built, so as to use static libs.
- make the config file derive defaults from the package.json file of the project.
- we add the templating of a number of additional files 

N.B. We should probably squash this PR before merging: the individual commits do not build (due to some git hackery on my part).


crates/ubrn_cli/src/android.rs
```diff
@@ -136,9 +148,74 @@ impl AndroidArgs {
...
+            Target::ArmeabiV7a => "armeabi-v7a",
+            Target::Arm64V8a => "arm64-v8a",
+            Target::X86 => "x86",
+            Target::X86_64 => "x86_64",
+        })
+    }
+}
+
+impl Target {
+    pub fn triple(&self) -> &'static str {
```
zzorba, line 213: Just for my own understanding, what is the history/implication of the name 'triple' here?

crates/ubrn_cli/src/codegen/templates/ModuleTemplate.java
```diff
@@ -0,0 +1,54 @@
...
+
+  static {
+    System.loadLibrary("{{ self.config.project.cpp_filename() }}");
+  }
+
+  private static native double nativeMultiply(double a, double b);
+  private static native boolean nativeInstallRustCrate(long rtPtr, boolean a);
+  private static native boolean nativeCleanupRustCrate(long rtPtr, boolean a);
+
+  @Override
```
zzorba, line 34: This is a placeholder right, eventually it gets deleted (maybe a TODO note is warranted)

crates/ubrn_cli/src/codegen/templates/NativeCodegenTemplate.ts
```diff
@@ -0,0 +1,11 @@
+// Generated by uniffi-bindgen-react-native
+import type { TurboModule } from 'react-native';
+import { TurboModuleRegistry } from 'react-native';
+
+export interface Spec extends TurboModule {
+  multiply(a: number, b: number): number;
```
zzorba, line 6: Same here, multiply is temporary right?

crates/ubrn_cli/src/codegen/templates/TurboModuleTemplate.cpp
```diff
@@ -0,0 +1,26 @@
...
+{%- let root = self.project_root() %}
+{%- let bindings = self.config.project.bindings.cpp_path(root) %}
+{%- let bindings = self.relative_to(root, bindings) %}
+{%- for m in self.config.modules %}
+#include "{{ bindings }}/{{ m.hpp_filename() }}"
+{%- endfor %}
+
+namespace matrixrustsdk {
+	using namespace facebook;
+	double multiply(double a, double b) {
```
zzorba, line 12: Note to remove down the road

crates/ubrn_cli/src/workspace.rs
```diff
@@ -6,15 +6,22 @@
...
+
+fn package_json_file() -> Result<Utf8PathBuf> {
     let pwd = ubrn_common::pwd()?;
 
     let package_json = ubrn_common::resolve(pwd, "package.json")?;
     Ok(package_json.expect("Must be run under a directory containing a package.json file"))
 }
 
+pub(crate) fn package_json() -> PackageJson {
+    let file = package_json_file().expect("Cannot file package.json");
```
zzorba, line 19: should this be "Cannot find package.json" instead of *file?

zzorba:
Change looks good, a few small questions/comments about placeholders.

👍 to squashing, cleaner git history + its nice if each commit builds (I always squash for this reason, since mistakes happen when developing and life is too short to spend time rewriting your git history on your branch more than you have to).

zzorba (approved):

crates/ubrn_cli/src/codegen/templates/NativeCodegenTemplate.ts
jhugman: yes, multiply is temporary. I should add some TODOs.

crates/ubrn_cli/src/codegen/templates/ModuleTemplate.java
jhugman: The `nativeInstallRustCrate` and `nativeCleanupRustCrate` is constant. The `multiply` is temporary, however.

crates/ubrn_cli/src/android.rs
jhugman: The target triple is the`cpu-vendor-os` that we're compiling for.

crates/ubrn_cli/src/workspace.rs
jhugman: The `expect` causes a `panic` if `package_json_file()` returns with an error.

PR #21, opened 2024-06-19
Add generated templates for iOS

jhugman:

This adds files specific to iOS: the `.mm`, `.h` and `.podspec`.

Because the podspec is likely to be evolve in the builder-bob template as react-native evolves, then we also add a `noOverwrite` to the project config file.


zzorba (approved):
LGTM

PR #23, opened 2024-07-01
Callbacks WIP

jhugman:
This is a work in progress PR, to implement callback interfaces from Rust back in to JS.

I have been thinking of this in terms of a U shaped bend.

1. At initialization, JS calls into C++ with a `vtable` object: an object with a callback per method in the interface. The code for this is in `CallbackInterfaceImpl.ts`.
2. The C++ then constructs a C struct (in `VTableStruct.cpp`) which contains one or more function pointers. The struct and function pointers header declarations are defined in the `callback_struct_decl` and `callback_fn_decl` template macros in `macros.cpp`. The function implementation is done in `VTableCallbackFunction.cpp`.
3. The C++ then passes the VTable struct in to Rust.
4. … time passes, the U bends.
5. …
6. Rust calls into the Rust trait implemented by uniffi. It calls into C++ via the function pointer.
7. C++ transforms the low level Rust structures in to low level JSI structures.
8. C++ calls the Typescript 
9. The typescript lifts the handle out of the handle map, and dispatches to the correct callback interface with the correct method.

Elided here for clarity:
5. an actual callback interface is lowered to a handle and the handle is passed via the normal serialization machinery; Rust has deserialization machinery to construct a Rust trait and its implementation.

Also elided:
10. The return and error status is caught in Typescript…
11. serialized and passed to C++…
12. and passed back to Rust.

I'm fairly good from 1 to 7. I am seeing problems with 8: calling JS from C++, which means that there is a problem with the code constructed at 2.

Here is the generated code for a no arg/no throw callback. The function pointer is type aliased to `UniffiCallbackInterfaceForeignGettersMethod5`, and each has its own namespace.

```cpp
// Callback function: UniffiCallbackInterfaceForeignGettersMethod5
//
// We have the following constraints:
// - we need to pass a function pointer to Rust.
// - we need a jsi::Runtime and jsi::Function to call into JS.
// - function pointers can't store state, so we can't use a lamda.
//
// For this, we store a lambda as a global, as `lambda`. The `callback` function
// calls the lambda, which itself calls the `body` which then calls into JS.
//
// We then give the `callback` function pointer to Rust which will call the
// lambda sometime in the future.
namespace uniffi_jsi::unifficallbackinterfaceforeigngettersmethod5 {
using namespace facebook;

// We need to store a lambda in a global so we can call it from
// a function pointer. The function pointer is passed to Rust.
static std::function<void(uint64_t, void *, RustCallStatus *)> lambda = nullptr;

// This is the main body of the callback. It's called from the lambda,
// which itself is called from the callback function which is passed to Rust.
static void body(jsi::Runtime &rt,
                 std::shared_ptr<react::CallInvoker> callInvoker,
                 jsi::Function &cb, uint64_t rs_uniffiHandle,
                 void *rs_uniffiOutReturn, RustCallStatus *uniffi_call_status) {

  // Convert the arguments from Rust, into jsi::Values.
  // We'll use the Bridging class to do this…
  auto js_uniffiHandle =
      uniffi_jsi::Bridging<uint64_t>::toJs(rt, rs_uniffiHandle);
  // … but we need to take extra care for the return value.
  // In the typescript we use a dummy object we called a ReferenceHolder.
  auto js_uniffiOutReturn =
      uniffi_jsi::Bridging<ReferenceHolder<void *>>::jsNew(rt);

  // The RustCallStatus is another very simple JS object which will
  // report errors back to Rust.
  auto uniffiCallStatus = uniffi_jsi::Bridging<RustCallStatus>::jsSuccess(rt);

  // Now we are ready to call the callback.
  // We should be using callInvoker at this point, but for now
  // we think that there are no threading issues to worry about.
  std::cout << "\tCalling cb.call() CallbackInterfaceForeignGettersMethod5"
            << std::endl;
  try {
    cb.call(rt, js_uniffiHandle, js_uniffiOutReturn, uniffiCallStatus);

    std::cout << "\tSuccess! cb.call() CallbackInterfaceForeignGettersMethod5"
              << std::endl;

    // Now copy the result back from JS into the RustCallStatus object.
    uniffi_jsi::Bridging<RustCallStatus>::copyFromJs(rt, uniffiCallStatus,
                                                     *uniffi_call_status);

  } catch (const jsi::JSError &error) {
    std::cout
        << "Error in callback UniffiCallbackInterfaceForeignGettersMethod5: "
        << error.what() << std::endl;
    uniffi_jsi::Bridging<RustCallStatus>::copyFromJs(rt, uniffiCallStatus,
                                                     *uniffi_call_status);
  }
}

static void callback(uint64_t rs_uniffiHandle, void *rs_uniffiOutReturn,
                     RustCallStatus *uniffi_call_status) {
  // The runtime, the actual callback jsi::funtion, and the callInvoker
  // are all in the lambda.
  std::cout << "C++: callback CallbackInterfaceForeignGettersMethod5 begins"
            << std::endl;
  lambda(rs_uniffiHandle, rs_uniffiOutReturn, uniffi_call_status);
  std::cout << "C++: callback CallbackInterfaceForeignGettersMethod5 ends"
            << std::endl;
}

static UniffiCallbackInterfaceForeignGettersMethod5
makeCallbackFunction(jsi::Runtime &rt,
                     std::shared_ptr<react::CallInvoker> callInvoker,
                     jsi::Function &cb) {
  std::cout
      << "makeCallbackFunction for CallbackInterfaceForeignGettersMethod5 START"
      << std::endl;
  std::thread::id thisThreadId = std::this_thread::get_id();
  lambda = [&rt, callInvoker, &cb,
            thisThreadId](uint64_t rs_uniffiHandle, void *rs_uniffiOutReturn,
                          RustCallStatus *uniffi_call_status) {
    if (std::this_thread::get_id() != thisThreadId) {
      std::cout << "C++: callback CallbackInterfaceForeignGettersMethod5 "
                   "running in a different thread"
                << std::endl;
    } else {
      std::cout << "C++: callback CallbackInterfaceForeignGettersMethod5 "
                   "running in the same thread as initialized"
                << std::endl;
    }
    body(rt, callInvoker, cb, rs_uniffiHandle, rs_uniffiOutReturn,
         uniffi_call_status);
  };
  std::cout
      << "makeCallbackFunction for CallbackInterfaceForeignGettersMethod5 END"
      << std::endl;
  return callback;
}
} // namespace uniffi_jsi::unifficallbackinterfaceforeigngettersmethod5
```


crates/ubrn_bindgen/src/bindings/react_native/gen_typescript/templates/macros.ts
```diff
@@ -110,7 +110,7 @@
 
 {%- macro call_async(obj_factory, callable) -%}
         await uniffiRustCallAsync({
```
zzorba, line 112: So I think this is declared to take a `() => bigint` but it seems like it is passing an object to it as the first arg?  Did I miss something?

zzorba: I misread the template, nevermind

crates/ubrn_bindgen/src/bindings/react_native/gen_typescript/compounds.rs
```diff
@@ -60,7 +60,7 @@ impl SequenceCodeType {
 
 impl CodeType for SequenceCodeType {
     fn type_label(&self, ci: &ComponentInterface) -> String {
-        format!("{}[]", CodeOracle.find(self.inner()).type_label(ci))
+        format!("Array<{}>", CodeOracle.find(self.inner()).type_label(ci))
```
jhugman, line 63: A small bugfix: `Array<Option<Int>` was being rendered as `number | undefined[]`.

crates/ubrn_bindgen/src/bindings/react_native/gen_typescript/templates/ErrorTemplate.ts
```diff
@@ -48,20 +49,23 @@ const {{ ffi_converter_name }} = (() => {
...
-            {%- endfor %} {
-                throw new UniffiInternalError.UnexpectedEnumCase();
+                case "{{ variant.name()|class_name(ci) }}": {
+                    intConverter.write({{ loop.index }}, into);
+                    break;
+                }
+            {%- endfor %}
+                default:{
+                    throw new UniffiInternalError.UnexpectedEnumCase();
+                }
```
jhugman, line 61: Changed all this because hermes object construction ends up with things like:

```ts
class SimpleError extends Error {}
class ErrorVariant extends SimpleError {}

const e = new ErrorVariant("foo");
if (e instanceof ErrorVariant) {
  console.log("What Spidermonkey says");
} else if (e instanceof Error) {
  console.log("What hermes says");
}
```

crates/ubrn_bindgen/src/bindings/react_native/gen_typescript/templates/InitializationTemplate.ts
```diff
@@ -0,0 +1,19 @@
...
+    const scaffoldingContractVersion = nativeModule().{{ ci.ffi_uniffi_contract_version().name() }}();
+    if (bindingsContractVersion != scaffoldingContractVersion) {
+        throw new Error("contractVersionMismatch");
+    }
+
+    {%- for (name, expected_checksum) in ci.iter_checksums() %}
+    if (nativeModule().{{ name }}() != {{ expected_checksum }}) {
+        throw new Error("apiChecksumMismatch: {{ name }}");
+    }
+    {%- endfor %}
```
jhugman, line 14: This needed to happen eventually; it just happened now.

crates/ubrn_bindgen/src/bindings/react_native/gen_typescript/templates/InitializationTemplate.ts
```diff
@@ -0,0 +1,19 @@
...
+
+    {% for fn in self.initialization_fns() -%}
+    {{ fn }}();
+    {% endfor -%}
```
jhugman, line 18: This was the real reason for this: at start-up, we register the callback vtable with C++ and then onwards to Rust. Only after that can user code use callbacks.

jhugman:
This is now working as expected, with tests passing. Wrapping the `callbackFunction` in a `std::shared_ptr<jsi::Object>` allows it to compile and run.

The remaining complexity was around: 

1. getting the correct `drop` call to be used, which meant generating a `UniffiCallbackInterfaceFree`  for each vtable.
2. finding out that the error types weren't being detected by `instanceof` which meant some re-writing of Errors. I'm a little annoyed that I hadn't caught this with other tests, so I added tests to `coverall`.

However, there is still some work to do:

- a dangling pointer assertion failure (see below) when Runtime is destroyed. This is almost certainly due to storing of `&rt` in the statically stored lambda, and if it is I am not sure what to do about it.
- if it isn't then I need to investigate. The other dangling pointer theory I have is it's the `&callbackFunction`.

```
Assertion failed: (((1u << 31) & refCount_) == 0 && "This PointerValue was left dangling after the Runtime was destroyed."), function invalidate, file hermes.cpp, line 916.
```

Either way, I think where we're at now with this PR is it's ready to review, and land before moving on to the above, and the `callInvokerWrapper`.

zzorba (approved):
LGTM, one possible issue with typescript I observed when running locally.

PR #24, opened 2024-07-03
Fix shared library handling for linux target.

wesc:
We previously handled a 'unix' target_os, however that's not a valid value ('unix' is a target family). The 'unix' target family would catch many platforms we don't want, so we specifically target linux.

See: https://stackoverflow.com/a/76480973


zzorba (approved):
LGTM

jhugman (approved):

PR #25, opened 2024-07-04
Use module destructor to prevent dangling pointers

jhugman:

This small PR adds a destructor to the Native module which nulls out the lambda which holds the dangling pointer to the `jsi::Runtime`.


zzorba (approved):
LGTM

PR #26, opened 2024-07-04
Wrap callInvoker in a UniffiCallInvoker

jhugman:
This PR introduces a `UniffiCallInvoker` to isolate the logic of calling functions that return a promise and don't return a promise, when called from Rust regardless of which thread it is executed on.

This PR only deals with calling JS functions that do not return a promise (i.e. non-async ones).


zzorba (approved):
LGTM

PR #28, opened 2024-07-10
Rewrite identifiers which also happen to be typescript keywords

jhugman:

This small PR detects if a `var_name` or `fn_name` collides with any reserved words; if it does, it rewrites it by adding a trailing underscore.


zzorba (approved):
LGTM

PR #30, opened 2024-07-10
Generate test files from library file, rather than UDL

jhugman:

This PR finishes off the `cargo xtask run` command for running test files.

As we start the newer fixtures and examples, we are no longer able to rely on UDL files: instead we have to use the compiled `.so`/`.dylib` files which include the whole `ComponentInterface` derived from the proc_macros as well as UDL files.


zzorba (approved):
LGTM

PR #31, opened 2024-07-15
Add event loop to test-runner

jhugman:

This PR adds a crude event loop to the test runner. Specifically:

- `setTimeout`/`clearTimeout`/`setImmediate`. This is largely derived from https://github.com/tmikov/hermes-jsi-demos/blob/master/evloop/evloop.cpp
- an implementation of the `ReactCommon::CallInvoker` interface
- an `asyncTest` method to the asserts library

The runner now runs until there are no more tasks left, either from `setTimeout` or
from in-flght callbacks on other threads.

This does not need to be more robust than that: any long-running Rust tasks should
protect the test-runner from exiting by setting a timeout in the test.

This is done implicitly by the `asyncTest` method.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -4,3 +4,4 @@ build/
 *.template.json
 *.md
 package.json
+*.js.inc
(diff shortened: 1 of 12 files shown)


zzorba (approved):
LGTM

PR #32, opened 2024-07-15
Calling Rust Futures from Javascript

jhugman:

This PR depends on #30 and #31, even though it may merge cleanly.

It adds the [`examples/futures` fixture](https://github.com/mozilla/uniffi-rs/tree/main/examples/futures) from `uniffi-rs`, and enough code to make it work. 

Specifically, it:

- Adds `callInvoker` to all `uniffi_jsi::Bridging` implementations.
- Fixes the TODO in the `async-rust-call.ts` file during the initial conversion from Swift to Typescript.
- Removes the rather hacky "StringLifter" which `rustCall` needed: this is now passed in for every call.


zzorba (approved):
LGTM

PR #33, opened 2024-07-18
Asynchronous Callbacks

jhugman:

This PR builds on the previous work on callbacks (#23), and promises (#32).

In uniffi parlance: 
- JS callbacks are _"foreign" traits_, 
- Asynchronous JS promises map to Rust _futures_,  so:
- Asynchronous JS callbacks are "foreign" futures.

As part of this work, 
- the `fixtures/futures` test fixture, and translates the tests
- additional test infrastructure to measure how long a call takes.
- the callinvoker is now passed to the `toJs` Bridging methods.

Some extra things to think about:

- errors are not being passed reliably back to the calling function. This is most likely due to "complex" errors that haven't yet been implemented. For that reason, I have left this work for a different PR. I have implemented the tests, but `x`'d them out.
- performance does not seem to be very fast for futures– the `alwaysReady` function can take around 100ms! I am fairly sure this is down to the test-runner naive implementations of the event loop and call invoker. I would like to measure this in an app as soon as possible.
- the concept of task cancellation is missing in Javascript, so this is difficult to test.


jhugman (commented):
Self-review

zzorba (approved):
LGTM

Sorry for delay on review

jhugman:
Thanks for the review @zzorba !

PR #34, opened 2024-07-19
Support released version of CallInvoker

jhugman:

This PR enables the `UniffiCallInvoker` to work with an older version of the `CallInvoker`, i.e. the version currently in release.


zzorba (approved):
LGTM

PR #35, opened 2024-07-22
Generate C++ files suitable for side-by-side compilation

jhugman:
According to [The Big O of Code Reviews](https://www.egorand.dev/the-big-o-of-code-reviews/), this is a O(_m_) change.

This includes several commits which were needed to get the `matrix-rust-sdk` generating and compiling.

The _m_ changes:

- Separate the C++ namespaces such that `uniffi_jsi` and `uniffi_runtime` are all in the `includes` directory, and the generated classes and functions are in their own namespaces. This is the largest commit in this PR.
- Conditionally compile the `registerNatives` method so that it is only available when run with the test-runner.
- Renaming the `CallFunc` type alias to `UniffiCallFunc` so as not to clash with the `facebook::react::CallFunc`.
- Removal of a redundant ForeignBytes function.


cpp/includes/RustBuffer.h
```diff
@@ -15,59 +15,3 @@ struct RustBuffer {
...
-  }
-
-  // If we want this to be used, we should make FfiType::requires_cleanup()
-  // return true.
-  static void argument_cleanup(jsi::Runtime &rt, RustBuffer buf) {
-    // NOOP
-  }
-};
-
-} // namespace uniffi_jsi
```
jhugman, line 73: This is moved into the `RustBufferHelper`: it calls into the Rust library in the `rustbuffer_free` and `rustbuffer_from_bytes`, so needs to be in each of the generated C++ files.

cpp/includes/UniffiRustCallStatus.h
```diff
@@ -11,67 +11,3 @@ struct RustCallStatus {
...
-          uniffi_jsi::Bridging<RustBuffer>::fromJs(rt, invoker, rbuf);
-    }
-    if (statusObject.hasProperty(rt, "code")) {
-      auto code = statusObject.getProperty(rt, "code");
-      status->code = uniffi_jsi::Bridging<uint8_t>::fromJs(rt, invoker, code);
-    }
-  }
-};
-
-} // namespace uniffi_jsi
```
jhugman, line 77: This moves into a `RustCallStatusHelper.cpp`, which is part of the generated file. This is because it calls into `Bridging<RustBuffer>`, which is now part of the generated file.

crates/ubrn_bindgen/src/bindings/react_native/gen_cpp/templates/StringHelper.cpp
```diff
@@ -3,15 +3,15 @@
...
 {% let func = ci.ffi_function_string_to_arraybuffer() %}
 {%- call cpp::cpp_fn_from_js_decl(func) %} {
-    return uniffi_jsi::Bridging<std::string>::string_to_arraybuffer(rt, args[0]);
+    return {{ ci.cpp_namespace_includes() }}::Bridging<std::string>::string_to_arraybuffer(rt, args[0]);
 }
 
 {% let func = ci.ffi_function_arraybuffer_to_string() %}
 {%- call cpp::cpp_fn_from_js_decl(func) %} {
-    return uniffi_jsi::Bridging<std::string>::arraybuffer_to_string(rt, args[0]);
+    return {{ ci.cpp_namespace_includes() }}::Bridging<std::string>::arraybuffer_to_string(rt, args[0]);
```
jhugman, line 16: We consolidate `uniffi_jsi` into a single method.

crates/ubrn_bindgen/src/bindings/react_native/mod.rs
```diff
@@ -378,6 +387,37 @@ impl FfiType {
     fn is_void(&self) -> bool {
         matches!(self, Self::VoidPointer)
     }
+
+    fn cpp_namespace(&self, ci: &ComponentInterface) -> String {
```
jhugman, line 391: This describes which  namespace the struct declaration is in.

The namespace where the `Bridging` class is in `bridging_namespace`.

jhugman (commented):
Self-review

zzorba (approved):
LGTM

PR #36, opened 2024-07-22
Delay callback initialization until after the module is loaded

jhugman:

This PR separates out the callback registration and checksum (in other backends this is called `uniffiEnsureInitialized`) with the module loading.

This was found in the differences between testing against hermes in a test harness and in a React Native setting.

It does this by changing the default export of the module from an object containing `FfiConverters` to one containing the the `initialize` function and the `converters`.

The `converters` import is changed to match this new export, and the `index.ts` template in the `generate turbo-module` job is also change to call the `initialize` functions.


zzorba (approved):
LGTM

PR #38, opened 2024-07-24
Implement non-flat errors

jhugman:

Rust does not have checked exceptions, instead uses a general purpose `Result<T, E>` enum.

Where `E` is a simple enum, uniffi calls this a "flat errror".

Where `E` to be an enum with associated values, uniffi calls this a "non-flat error".

This PR re-implements the generation of errors for typescript, so as to enable non-flat errors.

As part of this work: 

- implemented a collection of `Error` subclasses, one for each variant in the `Error` enum.
- implemented our own static `instanceOf` method for the (Rust) `Error` enums and (Typescript) `Error` subclasses.
- used these `instanceOf` in the callback and async-callback code to detect if the error was expected or not.
- enabled the skipped tests for futures fixture.
- added more tests to the coverall and callbacks fixture.
- renamed errors that are called `Error` (which collides with JS `Error`) to `Exception`. This tightens up the rule which was really only intended for JVM based languages.


zzorba (approved):
LGTM, great use of type assertions to build the custom instance_of

PR #41, opened 2024-07-25
Remove need for `enablePackageExports` in depending projects

### LINKED ISSUE #13: Use of package.exports requires experimental metro config
In order for metro to understand sub-package exports, users of this library have to enable the following in metro.config.js in the resolver config.

```
const {getDefaultConfig, mergeConfig} = require('@react-native/metro-config');

/**
 * Metro configuration
 * https://reactnative.dev/docs/metro
 *
 * @type {import('metro-config').MetroConfig}
 */
const config = {
  resolver: {
    unstable_enablePackageExports: true,
  },
};

module.exports = mergeConfig(getDefaultConfig(__dirname), config);
```

In order to avoid this until it becomes non-experimental, it would be good to move to a flat package structure with a single 'main'.

I think we probably also want the package to compile to *.js/*.d.ts files as in https://github.com/jhugman/uniffi-bindgen-react-native/pull/12, though the build is not correctly configured yet.

jhugman:
Fixes #13.

### DIFF
--- a/crates/ubrn_bindgen/src/bindings/react_native/gen_typescript/mod.rs
+++ b/crates/ubrn_bindgen/src/bindings/react_native/gen_typescript/mod.rs
@@ -162,17 +162,17 @@ impl<'a> TypeRenderer<'a> {
     // ```
     //
     // Returns an empty string so that it can be used inside an askama `{{ }}` block.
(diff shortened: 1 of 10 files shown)


zzorba (approved):
LGTM

PR #42, opened 2024-07-25
Use node to resolve path to ubrn module.

wesc:
This is a similar change to what we had to do in the CMakeList generation.

### DIFF
--- a/bin/cli
+++ b/bin/cli
@@ -5,16 +5,15 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/
 #
 set -e
-script_dir="$(dirname "${BASH_SOURCE[0]}")"
-if [[ "$script_dir" == *".bin" ]] ; then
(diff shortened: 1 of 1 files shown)


bin/cli.sh
```diff
@@ -5,16 +5,14 @@
...
-    exit 1
-fi
 
+# Use Node to resolve the path to the module. The require.resolve
+# function returns a thing like:
+#
+# PATH_TO/node_modules/uniffi-bindgen-react-native/dist/objects.js
+#
+# So we drop the /dist and everything after it.
+root_dir=$(node --print "require.resolve('uniffi-bindgen-react-native/objects')" | sed 's/\/dist.*//')
```
zzorba, line 15: I think this should just be uniffi-bindgen-react-naive and it needs to strip out `typescript/src/index\\.ts` instead of dist.

zzorba (approved):
LGTM

PR #43, opened 2024-07-25
Strip extension of cli script

jhugman:
When used from within a package.json `scripts` script, yarn tries to work out what to do with it based on the extension. It doesn't recognize `.sh` extension.

This commit removes the `.sh` extension.


wesc (approved):

wesc:
Dunno where it is in the commit but file permission should be executable, right?

jhugman:
> Dunno where it is in the commit but file permission should be executable, right?

Yes, it'll be in the initial commit that introduced the file.

PR #45, opened 2024-07-25
Prefer interface types to concrete class names

jhugman:
When defining a new class, uniffi-bindgen-react-native routinely generates an interface.

By convention, we're calling this `FooInterface`.

```ts
interface FooInterface {
    method(obj: Bar): Baz
}

class Foo implements FooInterface {
    method(obj: Bar): Baz
}
```

When a method on a `FooInterface` returns another object, or accepts an object as an argument— say `Bar`— prior to this commit, both the `FooInterface` and the concrete `Foo` type would use the concrete `Bar`.

```ts
interface FooInterface {
    method(obj: Bar): Baz
}

class Foo implements FooInterface {
    method(obj: Bar): Baz {
        // …
    }
}
```

This changes this, to use the interface name.

```ts
interface FooInterface {
    method(obj: BarInterface): BazInterface
}

class Foo implements FooInterface {
    method(obj: BarInterface): BazInterface {
        // …
    }
}
```

This makes it easy for creating mocks and dummy interfaces.

Question:

I notice that the `type` and the `class` namespaces seem to be disjoint, so this is valid:

```ts
interface Foo {
  method(bool: boolean): void;
}

class Foo implements Foo {
  method(bool: boolean): void {}
}

class Bar implements Foo {
  method(bool: boolean): void {}
}

const foo: Foo = new Bar();
```

Should I make these the same names? If keeping them different, what should they be?

If they are different, and we want to keep this commit, then other questions need to be thought about:

- what should the behaviour of alternate constructors be? (currently, they return the interface, not the concrete type)
- how should we protect the object `FfiConverter` which now accepts an interface, but can only handle a concrete type? Does it really matter, given that it will only cause a failing test?

/cc @zzorba WDYT?

### DIFF
--- a/crates/ubrn_bindgen/src/bindings/react_native/gen_typescript/callback_interface.rs
+++ b/crates/ubrn_bindgen/src/bindings/react_native/gen_typescript/callback_interface.rs
@@ -19,6 +19,10 @@ impl CallbackInterfaceCodeType {
 }
 
 impl CodeType for CallbackInterfaceCodeType {
+    fn decl_type_label(&self, ci: &ComponentInterface) -> String {
+        format!("{}Impl", CodeOracle.class_name(ci, &self.id))
+    }
+
     fn type_label(&self, ci: &ComponentInterface) -> String {
         CodeOracle.class_name(ci, &self.id)
     }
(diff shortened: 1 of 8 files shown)


zzorba:
So change looks good. 

I think if there are situations where we NEED the object and the interface is insufficient, we should probably keep them as separate names (prefix I, suffix Type, something like that).

If we can always just refer to them through the interface, then I think the same name is fine (though by no means required, feels like the internet is of two minds as to whether doing the same name thing is cool or bad)

jhugman:
I've kept the `Interface` split here because I couldn't implement the interface without ts wanting me to extend it.

zzorba (approved):
LGTM

PR #10, opened 2024-05-29
Draft: Hooking Hermes GC to Rust (do not merge)

jhugman:
This changes the implementation of `bless` to a HostObject.

Ultimately, this commit may not make it to main, as it doesn't yet work.

PR #46, opened 2024-07-30
Add hooks into HostObject destructors

jhugman:
Fixes #10 .

This PR adds C++ destructor hooks from uniffi `interface` objects, to call the corresponding Rust `drop` methods.

The impedance mismatch between a GC'd language and an [RAII](https://en.wikipedia.org/wiki/Resource_acquisition_is_initialization) language means that one is a leaky abstraction for the other.

From https://github.com/facebook/hermes/issues/982#issuecomment-1771325667:

> Generally speaking, when dealing with native resources, it is
recommended to free the native resource explicitly when you are done
with it, because garbage collection is unpredictable and should not be
relied on for freeing native resources timely."

From the HostObject's destructor docs:

> The C++ object's dtor will be called when the GC finalizes this
object.  (This may be as late as when the Runtime is shut down.)
You have no control over which thread it is called on.  This will
be called from inside the GC…

For this reason, the `uniffiDestroy` method is left in place for now: if you absolutely need to drop a Rust peer object, then you should call the `uniffiDestroy` method of the correspond JS object.

From the integration tests alone, it is difficult to show that this is working; they are too short-lived for the Runtime to be still running and garbage collection to have taken place.

However, the `test_callbacks.ts` did cause an ungraceful shutdown when the destructor called into Rust which then called back into JS– which had already been killed.

The situation will only really get better when React Native/Hermes [supports a robust `FinalizationRegistry` implementation](https://github.com/facebook/hermes/issues/1440).


crates/ubrn_bindgen/src/bindings/react_native/gen_cpp/templates/VTableCallbackFunction.cpp
```diff
@@ -117,6 +117,19 @@ namespace {{ ns }} {
...
+        // call into Javascript. BUT how do we tell if the runtime has shutdown?
+        //
+        // Answer: the module destructor calls into callback `cleanup` method,
+        // which nulls out the rsLamda.
+        //
+        // If rsLamda is null, then there is no runtime to call into.
+        if (rsLambda == nullptr) {
+            // This only occurs when destructors are calling into Rust free/drop,
+            // which causes the JS callback to be dropped.
+            return;
```
zzorba, line 130: clever

zzorba (approved):
LGTM

PR #48, opened 2024-07-31
Enums with associated values and named associated values

jhugman:
This changes the way enums with associated values (a.k.a. tagged unions) are represented in typescript.

Previously, they were raw objects, now they are classes.

This allows us to add extra methods, including handy constructors and `instanceOf` methods.

The interface for the `MyEnum` type is approximately `{ tag: MyEnum_tags, data: MyEnum_interface }`, where `MyEnum_tags` is exported.

The tag allows us to to interesting pattern matching like switch statements.

e.g. consider an enum for `Sheep`, `Goat(String, u32)`, `Cat { owner: Option<String>, name: String }`

The values are constructed like so:

```typescript
const sheep = new Animal.Sheep();
const jeff = new Animal.Goat("Jeff", 17);
const roberta = new Animal.Cat({ name: "Roberta", owner: undefined });
```

For pattern matching, a simple tag enum is generated and exposed:

```typscript
enum Animal_tags {
    Sheep, Goat, Cat
}
```

Typescript will infer the correct types of the `value.data` property:

```typescript
function name(animal: Animal): string {
    switch (animal.tag) {
        case Animal_tags.Sheep:
            return "mah";
        case Animal_tags.Goat: {
            const [name, age] = animal.data;
            return name;
        }
        case Animal_tags.Cat: {
            const {name, owner} = animal.data;
            return name;
        }
        default:
            throw new Error("Unreachable");
    }
}
````

For convenience, a `new` method is also generated:

```typscript
const berry = Animal.Cat.new({ name: "Berry", owner: undefined });
````

Having `new` as a method means that changing `Animal` from a flat enum (i.e. one without associated types) to one that does means only changing one side of the literal.

Also: `instanceOf` methods:

```typscript
const ermintrude: any = Animal.Goat.new("Ermintrude", 18);

if (Animal.instanceOf(ermintrude)) {
    // ermintrude is an animal
}
if (Animal.Goat.instanceOf(ermintrude)) {
    // ermintrude is goat.
}
```

### DIFF
--- a/crates/ubrn_bindgen/src/bindings/react_native/gen_typescript/filters.rs
+++ b/crates/ubrn_bindgen/src/bindings/react_native/gen_typescript/filters.rs
@@ -108,7 +108,33 @@ pub fn variant_discr_literal(
     ci: &ComponentInterface,
 ) -> Result<String, askama::Error> {
     let literal = e.variant_discr(*index).expect("invalid index");
-    Ok(Type::Int32.as_codetype().literal(&literal, ci))
+    let ts_literal = Type::Int32.as_codetype().literal(&literal, ci);
+    Ok(match literal {
+        Literal::String(_) => ts_literal,
+        Literal::UInt(_, _, typ) | Literal::Int(_, _, typ)
+            if !matches!(&typ, &Type::Int64 | &Type::UInt64) =>
+        {
+            ts_literal
+        }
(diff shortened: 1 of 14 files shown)


zzorba (approved):
LGTM, love this new syntax

PR #50, opened 2024-07-31
Rename RecordFactory to Record

jhugman:
We've been able to declare a type and a utility object of the same name.

Before this change, we could do:

```typescript
const v: MyRecord = MyRecordFactory.create({ foo: 1 });
```

Now we can do:

```typescript
const v: MyRecord = MyRecord.create({ foo: 1 });
```

This reduces the number of specially generated names which are really implementation details of uniffi.


zzorba (approved):
LGTM

PR #51, opened 2024-07-31
Implement UniffiTraits for Objects

jhugman:
UniffiTraits are a handful of utility traits that Rust can implement automatically.

If a UniffiTrait is implemented, then the expected behaviour is that the foreign language forwards calls on to Rust.

e.g. adding `#[uniffi::export(Display)]` causes a `toString()` method to Javascript.

Unfortunately, the corresponding methods in Javascript are all non-standard.

I feel sure that there `valueOf()` and `[Symbol.toPrimitive]` may be useful to implement here, but I don't know enough about the internals to be sure this is the right thing to do.


zzorba (approved):
LGTM

PR #53, opened 2024-08-02
Implementing custom types

jhugman:

This adds the `custom-types` example fixture allowing you to write handwritten conversion logic into the generated typescript.

To enable this, we add a:
- `--toml` argument to the `cargo xtask run` command line for testing, 
- `--config` argument to the `ubrn generate bindings`; this is to keep it the same as other uniffi-bindgen backends,
- `uniffiToml` property in bindings section of the `config.yaml` files used elsewhere.


zzorba (approved):
LGTM

PR #54, opened 2024-08-05
Implement objects thrown as errors

jhugman:
This PR adds the error-types fixture.

It:

- introduces a `UniffiThrownObject<T>` wrapper error, which holds the object from lifted Rust.
- refactors ordinary non-flat errors to use the same template as non-flat enums.
  - this means that errors have `tag` and `data` properties, with `data` holding any parameters in a non-flat error.
- discovers that in hermes, subclasses of errors can have properties and static methods, but not calculated properties or methods.


fixtures/error-types/tests/bindings/test.kts
```diff
@@ -0,0 +1,113 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import uniffi.error_types.*;
+import kotlinx.coroutines.*
+
+try {
```
zzorba, line 8: Are these outputs from stock uniffi, not uniffi-bindgen?

zzorba (approved):
LGTM

fixtures/error-types/tests/bindings/test.kts
jhugman: Yes, they're left in error. Removed.

PR #56, opened 2024-08-06
Tighten up futures testing

jhugman:
This PR makes the `test_futures.ts` tests run in series.

This makes it easier to test/demonstrate that:

- the `async-callbacks` are cleaned up (so we can use this to keep the promise in memory, and have a future place for an AbortController to live)
- the `async-rust-call` are cleaned up.

Additional tests/documentation is also put in place for a future implementation of task cancellation, both Rust cancelling JS and the other way round.

This hasn't been implemented here: 

- for Rust cancelling JS: `AbortController` exists, but doesn't seem to be widely adopted. It is not yet implemented by Hermes.
- for JS cancelling Rust: an FFI method can be easily exposed, but it is not clear what the most JS-ergonomic way to do this might be. 
- for Rust cancelling Rust, and it propagating to JS: I'm not 100% sure, but I don't think it is currently supported by uniffi-rs.
  - There seems to be missing tests for this case: [Swift test](https://github.com/mozilla/uniffi-rs/blob/main/fixtures/futures/tests/bindings/test_futures.swift#L401-L429), [Kotlin test](https://github.com/mozilla/uniffi-rs/blob/main/fixtures/futures/tests/bindings/test_futures.kts#L325-L352). Missing is the case where `timeoutMs` < `releaseAfterMs`, such that the Rust aborts after the timeout.
  - Only the Swift implementation makes mention of [`CALL_CANCELLED`, and `fatalError`s on detection](https://github.com/mozilla/uniffi-rs/blob/7ff7584ce06a2d5160d598b319689d0f691c9ddf/uniffi_bindgen/src/bindings/swift/templates/Helpers.swift#L104-L105).

### DIFF
--- a/crates/ubrn_bindgen/src/bindings/react_native/gen_cpp/templates/macros.cpp
+++ b/crates/ubrn_bindgen/src/bindings/react_native/gen_cpp/templates/macros.cpp
@@ -33,7 +33,7 @@ jsi::Value {{ module_name }}::{% call cpp_func_name(func) %}(jsi::Runtime& rt, c
 
 {%- macro cpp_fn_rust_caller_body_with_func_name(func, func_name) %}
         {%- if func.has_rust_call_status_arg() %}
-        RustCallStatus status = { 0 };
+        RustCallStatus status = {{ ci.cpp_namespace() }}::Bridging<RustCallStatus>::rustSuccess(rt);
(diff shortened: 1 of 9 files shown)


zzorba (approved):
LGTM

PR #57, opened 2024-08-06
Use DataView instead of TypedArrays in RustBuffer

jhugman:
Using typed arrays had a number of disadvantages, most notably:

- Handling endian-ness was a manual process which was non-obvious.
- The offset rules, i.e. you can't write to the underlying array buffer at offsets that were not multiples of the element byte length.

This resulted in lots of copying. For example, when writing a `BigInt` to the byte offset 11 a destination `ArrayBuffer`:

1. an transitional `ArrayBuffer` of 8 bytes was created
2. a `Uint64Array` typed array was created to view that transitional ArrayBuffer.
3. the `BigInt` was written into the `Uint64Array`, i.e. the transitional array.
4. the transitional array was reversed, because of endian-ness differences between Rust and Hermes.
5. a `Uint8Array` view was created for each of the transitional array and the destination array.
6. the transitional array was copied into the destination array.

This was a lot of work for a single write: steps 1, 3, 4 and 6 involve `O(n)` work.

This PR replaces the typed arrays with a `DataView` which is a lot simpler:

1. a `DataView` of the destination array is created
2. the `BigInt` is written into the `DataView`, of the correct endianness.

Only step 2 involves `O(n)` work.

The work allowed us to simplify the `RustBuffer`: the `read` and `write` method are both replaced by two simpler methods:

- `readWithView` and `readBytes`.
- `writeWithView` and `writeBytes`.

### DIFF
--- a/crates/ubrn_bindgen/src/bindings/react_native/gen_typescript/templates/StringHelper.ts
+++ b/crates/ubrn_bindgen/src/bindings/react_native/gen_typescript/templates/StringHelper.ts
@@ -24,13 +24,14 @@ const FfiConverterString = (() => {
         }
         read(from: RustBuffer): TypeName {
             const length = lengthConverter.read(from);
-            return from.read(length, arrayBufferToString);
+            const bytes = from.readBytes(length);
+            return arrayBufferToString(bytes);
         }
         write(value: TypeName, into: RustBuffer): void {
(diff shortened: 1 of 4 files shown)


zzorba (approved):
Amazing, much simpler

PR #58, opened 2024-08-06
Remove deadcode

jhugman:

- Removed dead rust code.
- Simplified template Typescript code: most removing empty or almost redundant files.
- Removed dead code in `typescript/src` files.
- Removing spurious python, swift and kotlin code.
- Removed dead code in C++ templates and `cpp/includes`.
- Renamed `VTableStruct.cpp` and `VTableCallbackFunction.cpp` to drop misleading `VTable` prefix.
- Removed duplicated license headers.


zzorba (approved):
LGTM

PR #59, opened 2024-08-07
Rename `data` property to `inner` in enums and errors

jhugman:

This mechanical rename changes the type of enums and errors from:

```ts
type MyEnum = {
  tag: MyEnum_Tags,
  data: MyEnum_data,
};
```
to:
```ts
type MyEnum = {
  tag: MyEnum_Tags,
  inner: MyEnum_data,
};
```

This is to match the `getInner` and `hasInner` static methods for errors.


zzorba (approved):
I like it, clearer name

PR #62, opened 2024-08-07
Adding a test around explicit library-specific cancellation

jhugman:

This PR changes a test to catch an error.

I thought that we should be expecting a `CancellationError`, but after speaking to the uniffi-rs team, it turned out I should have been expecting an `AsyncError.Timeout`.

This test re-writes the test so `Timeout` is caught.


zzorba (approved):
LGTM

PR #63, opened 2024-08-14
Initial commit of mdbook

jhugman:
This PR should generate an mdbook from the documents in `docs`.

It is only the infrastructure to install and serve the static site.


zzorba (approved):
LGTM

PR #66, opened 2024-08-16
Add type collection pass to get more info about External types

jhugman:

This is a large diff, but most of the work is mechanical.

1. Allow the test-runner to initialize more than one C++ file. 
2. Adding a lookup for `External` types so the templates have more detail about what the `External` (imported) types actually are.
3. Adding a `ext-types` fixture.

The most interesting of these is the second:

Before rendering any files, a pass is taken of all `ComponentInterface`s.

Each type is collected, and added to a type map, keyed by module-path and name.

When an external type is encountered later, the type map is consulted, and the actual type
is used.

This is to remove a deficiency in `Type::External`: the `ExternalKind` enum did not map completely onto a `Type`:
before this change, it wasn't clear what sort of type was being imported, and so it was difficult/impossible
to work out __how__ to import (`import { X }` vs `import { type X  }`), or __what__ to import: `X` vs `XInterface`.

This meant changing the `type_name`, `ffi_converter_name` and some other filter methods, necessitating __many__ template files. These changes can be largely skimmed over.

The test also caught a runtime bug, which is also fixed in this PR: when a Rust trait is implemented by Rust, but exposed (as a typescript `interface`), this was erroring at runtime with a `UniffiStaleHandleError`. This is now fixed.


zzorba (approved):
LGTM

PR #70, opened 2024-08-17
Allow enum variant names that collide with existing types

jhugman:

This small PR tweaks the Variant classes for our typescript version of Tagged Enums, to allow Variants to have names of existing types.

For example, it's not uncommon for Rust programs to have enums constructed like so:

```rs
struct Dog {}
struct Cat {}

enum Animal {
  Dog(Dog),
  Cat(Cat),
}
```

The way we've modeled tagged enums are with a frozen object containing a class for each Variant.

So this would translate to:

```typescript
type Dog = …; // Dog defined elsewhere
type Cat = …;

const Animal = (() => {
    class Dog {
        constructor(public value: Dog) {
            // This Dog value is the wrong type!
            // value is the type of the Variant class.
        }
    }
    class Cat {
        constructor(public value: Cat) {}
    }
    return Object.freeze({ Dog, Cat });
})();
```

This commit changes the naming of the Variant class:

```typescript
type Dog = …; // Dog defined elsewhere
type Cat = …;

const Animal = (() => {
    class Dog_ {
        constructor(public value: Dog) {
            // There is no variant class called Dog yet, just Dog_,
            // so value is the correct type.
        }
    }
    class Cat_ {
        constructor(public value: Cat) {}
    }
    // We rename the Dog_ variant class as Dog here,
    // so the only way to access this class is via Animal.Dog.
    return Object.freeze({ Dog: Dog_, Cat: Cat_ });
})();
```

### DIFF
--- a/crates/ubrn_bindgen/src/bindings/react_native/gen_typescript/templates/TaggedEnumTemplate.ts
+++ b/crates/ubrn_bindgen/src/bindings/react_native/gen_typescript/templates/TaggedEnumTemplate.ts
@@ -11,10 +11,23 @@ export enum {{ kind_type_name }} {
 {%- call ts::docstring(e, 0) %}
 export const {{ decl_type_name }} = (() => {
   {%- for variant in e.variants() %}
-    {%- let variant_name = variant.name()|class_name(ci) %}
+    {#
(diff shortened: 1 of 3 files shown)


zzorba (approved):
LGTM

PR #71, opened 2024-08-17
Break up the ubrn command line build commands

jhugman:

This adds a number of suppression flags to the command line. This is to make it:

- easier to generate code using your own build process
- easier to customize parts of the build process
- cache as much as possible of the previous build

For `build ios`:

- `--no-xcodebuild` suppress the creation of an xcframework. The xcframework is expected, but with this flag, you are expected to provide it; perhaps after generating swift from a library file. (future work idea: add an arg to generate Swift from uniffi).
- `--no-cargo` detects if any targets are already built, and uses that to generate library files. If no targets are built, then build all, re-using the  library files next time.

For `build android`

- `--no-jniLibs`, this is Android reflection of `--no-xcodebuild`. I'm not sure what this would be used for.
- `--no-cargo` detects if any targets are already built, and uses that to generate library files. If no targets are built, then build all, re-using the  library files next time.

A new command: `generate all --config config.yaml lib.a`:

- This combines `generate bindings` and `generate turbo-module` into a higher level command which accepts a config file and a pre-built library file.

### DIFF
--- a/crates/ubrn_cli/src/android.rs
+++ b/crates/ubrn_cli/src/android.rs
@@ -262,6 +262,10 @@ impl AndroidArgs {
     pub(crate) fn project_config(&self) -> Result<ProjectConfig> {
         self.config.clone().try_into()
     }
+
+    pub(crate) fn config(&self) -> Utf8PathBuf {
+        self.config.clone()
+    }
 }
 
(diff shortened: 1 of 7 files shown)


zzorba (approved):
LGTM, appreciate the increased granularity here.

Issue #75, opened 2024-08-19
Warnings about private fields being exposed on generated types


zzorba:
Looks like with recent code, typescript is complaining about some fields getting exposed.

Doing edits locally, changing the macro of macros.ts to the following:
```
    typeof {{ decl_type_name }}[keyof Omit<typeof {{ decl_type_name }}, 'instanceOf' | '__uniffiTypeName' | '__variant' | '__variantName'>]

```
Seemed to fix the problem.

zzorba:
@jhugman this is the issue we discussed, I do think moving these off the super class is the right course of action.

I probably won't have time to work on it for the next couple weeks, so we can see who gets to it first.

PR #76, opened 2024-08-19
Fix macro for inherited types to avoid typescript error

zzorba:
I believe this will address https://github.com/jhugman/uniffi-bindgen-react-native/issues/75

### DIFF
--- a/typescript/src/enums.ts
+++ b/typescript/src/enums.ts
@@ -5,8 +5,20 @@
  */
 export class UniffiEnum {
   constructor(
-    private readonly __uniffiTypeName: string,
-    private readonly __variantName: string,
-    private readonly __variant: number,
+    /*
+     * This member should be private, but typescript requires
+     * it be public because it cannot enforce it.
+     */
+    public readonly __uniffiTypeName: string,
+    /*
+     * This member should be private, but typescript requires
+     * it be public because it cannot enforce it.
+     */
+    public readonly __variantName: string,
+    /*
+     * This member should be private, but typescript requires
+     * it be public because it cannot enforce it.
+     */
+    public readonly __variant: number
   ) {}
 }
--- a/typescript/src/errors.ts
+++ b/typescript/src/errors.ts
@@ -11,10 +11,22 @@
 // for each error.
 export class UniffiError extends Error {
   constructor(
-    private readonly __uniffiTypeName: string,
-    private readonly __variantName: string,
-    private readonly __variant: number,
-    message?: string,
+    /*
+     * This member should be private, but typescript requires
+     * it be public because it cannot enforce it.
+     */
+    public readonly __uniffiTypeName: string,
+    /*
+     * This member should be private, but typescript requires
+     * it be public because it cannot enforce it.
+     */
+    public readonly __variantName: string,
+    /*
+     * This member should be private, but typescript requires
+     * it be public because it cannot enforce it.
+     */
+    public readonly __variant: number,
+    message?: string
   ) {
     // We append the error type and variant to the message because we cannot override `toString()`—
     // in errors.test.ts, we see that the overridden `toString()` method is not called.
@@ -26,7 +38,7 @@ export class UniffiError extends Error {
     return UniffiError.createMessage(
       this.__uniffiTypeName,
       this.__variantName,
-      this.message,
+      this.message
     );
   }
 
@@ -37,7 +49,7 @@ export class UniffiError extends Error {
   private static createMessage(
     typeName: string,
     variantName: string,
-    message: string | undefined,
+    message: string | undefined
   ): string {
     const prefix = `${typeName}.${variantName}`;
     if (message) {
@@ -54,7 +66,7 @@ export class UniffiThrownObject<T> extends Error {
   constructor(
     private readonly __uniffiTypeName: string,
     public readonly inner: T,
-    message?: string,
+    message?: string
   ) {
     // We append the error type and variant to the message because we cannot override `toString()`—
     // in errors.test.ts, we see that the overridden `toString()` method is not called.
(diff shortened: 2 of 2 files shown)


zzorba:
Basically our exporting trick has one problem, https://github.com/microsoft/TypeScript/issues/30355

So I marked them public and put a strongly worded note on the fields in question.  We could also suppress the typescript warning, but it has to be done at each export site (annoying)

typescript/src/enums.ts
```diff
@@ -5,8 +5,20 @@
...
+    /*
+     * This member should be private, but typescript requires
+     * it be public because it cannot enforce it.
+     */
+    public readonly __variantName: string,
+    /*
+     * This member should be private, but typescript requires
+     * it be public because it cannot enforce it.
+     */
+    public readonly __variant: number
```
jhugman, line 22: Nit: prettier wants a trailing comma.

typescript/src/errors.ts
```diff
@@ -11,10 +11,22 @@
...
+     * This member should be private, but typescript requires
+     * it be public because it cannot enforce it.
+     */
+    public readonly __variantName: string,
+    /*
+     * This member should be private, but typescript requires
+     * it be public because it cannot enforce it.
+     */
+    public readonly __variant: number,
+    message?: string
```
jhugman, line 29: Nit: prettier wants a trailing comma.

typescript/src/errors.ts
```diff
@@ -26,7 +38,7 @@ export class UniffiError extends Error {
     return UniffiError.createMessage(
       this.__uniffiTypeName,
       this.__variantName,
-      this.message,
+      this.message
```
jhugman, line 41: Nit: prettier wants a trailing comma.

typescript/src/errors.ts
```diff
@@ -54,7 +66,7 @@ export class UniffiThrownObject<T> extends Error {
   constructor(
     private readonly __uniffiTypeName: string,
     public readonly inner: T,
-    message?: string,
+    message?: string
```
jhugman, line 69: Nit: prettier wants a trailing comma.

typescript/src/errors.ts
```diff
@@ -66,7 +78,7 @@ export class UniffiThrownObject<T> extends Error {
     return UniffiThrownObject.createMessage(
       this.__uniffiTypeName,
       this.inner,
-      this.message,
+      this.message
```
jhugman, line 81: Nit: prettier wants a trailing comma.

typescript/src/errors.ts
```diff
@@ -81,7 +93,7 @@ export class UniffiThrownObject<T> extends Error {
   private static createMessage<T>(
     typeName: string,
     obj: any,
-    message: string | undefined,
+    message: string | undefined
```
jhugman, line 96: Nit: prettier wants a trailing comma.

typescript/src/errors.ts
```diff
@@ -108,7 +120,7 @@ export const UniffiInternalError = (() => {
   class BufferOverflow extends Error {
     constructor() {
       super(
-        "Reading the requested value would read past the end of the buffer",
+        "Reading the requested value would read past the end of the buffer"
```
jhugman, line 123: Nit: prettier wants a trailing comma.

typescript/src/errors.ts
```diff
@@ -145,14 +157,14 @@ export const UniffiInternalError = (() => {
   class ContractVersionMismatch extends Error {
     constructor(rustVersion: any, bindingsVersion: any) {
       super(
-        `Incompatible versions of uniffi were used to build the JS ($${bindingsVersion}) from the Rust (${rustVersion})`,
+        `Incompatible versions of uniffi were used to build the JS ($${bindingsVersion}) from the Rust (${rustVersion})`
```
jhugman, line 160: Nit: prettier wants a trailing comma.

typescript/src/errors.ts
```diff
@@ -145,14 +157,14 @@ export const UniffiInternalError = (() => {
...
       );
     }
   }
   class ApiChecksumMismatch extends Error {
     constructor(func: string) {
       super(
-        `FFI function ${func} has a checksum mismatch; this may signify previously undetected incompatible Uniffi versions`,
+        `FFI function ${func} has a checksum mismatch; this may signify previously undetected incompatible Uniffi versions`
```
jhugman, line 167: Nit: prettier wants a trailing comma.

typescript/src/errors.ts
```diff
@@ -37,7 +49,7 @@ export class UniffiError extends Error {
   private static createMessage(
     typeName: string,
     variantName: string,
-    message: string | undefined,
+    message: string | undefined
```
jhugman, line 52: Nit: prettier wants a trailing comma.

typescript/src/errors.ts
```diff
@@ -11,10 +11,22 @@
...
+    /*
+     * This member should be private, but typescript requires
+     * it be public because it cannot enforce it.
+     */
+    public readonly __variantName: string,
+    /*
+     * This member should be private, but typescript requires
+     * it be public because it cannot enforce it.
+     */
+    public readonly __variant: number,
```
jhugman, line 28: I wonder if this would be better as private properties in the generated subclasses rather than the super class.

My hunch is that the DX of having `__` properties turn up in autocomplete becomes annoying after a while.

WDYT? Is that annoyance a non-issue, or is it worth the extra work?

jhugman (approved):
I've added some nits and a question, but nothing that should warrant a second review cycle. Good job! Thanks for stepping in for this.

typescript/src/errors.ts
zzorba: Yes, that would be better, and they should still be accessible on the union type we're creating. AFAICT it is just the inheritence + const composure that is the problem.

It would be very nice if the __ didn't show up in autocomplete, since you aren't supposed to use them -- makes the generated code a lot more self documenting.

jhugman:
Closing this in favour of #84.

PR #82, opened 2024-08-30
Add fixture for timestamp and duration

jhugman:

This PR tests roundtripping and overflows of the `timestamp` and `duration` uniffi types.

These were added to uniffi-rs shortly before custom types were implemented, and arguably should be implemented as custom types.

| udl | Rust | Javascript | Comments |
| - | - | - | - |
| `timestamp` | `std::time::SystemTime` | `Date` | JS only supports dates up to [8.64e15 ms either side of the epoch](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date#the_epoch_timestamps_and_invalid_date). Uniffi will throw an error if the date is outside of these times  |
| `duration` | `chrono::Duration` | `number` |

Javascript represents numbers as 64 bit floating point numbers. The FFI protocol represents timestamps as `BigInt(seconds)` + `Number(nanoseconds)`. This  conversion will lead to inaccuracies at large numbers or high precision numbers.

### DIFF
--- a/typescript/src/ffi-converters.ts
+++ b/typescript/src/ffi-converters.ts
@@ -169,11 +169,11 @@ export const FfiConverterDuration = (() => {
       if (ms === Number.POSITIVE_INFINITY || ms === Number.NEGATIVE_INFINITY) {
         throw new UniffiInternalError.NumberOverflow();
       }
-      return ms + nanos / 1000;
+      return ms + nanos / nanosPerMs;
     }
     write(value: UniffiDuration, into: RustBuffer): void {
       const ms = value.valueOf();
-      const secsBigInt = BigInt(ms) / msPerSecBigInt;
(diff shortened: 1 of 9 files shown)


typescript/src/ffi-converters.ts
```diff
@@ -169,11 +169,11 @@ export const FfiConverterDuration = (() => {
       if (ms === Number.POSITIVE_INFINITY || ms === Number.NEGATIVE_INFINITY) {
         throw new UniffiInternalError.NumberOverflow();
       }
-      return ms + nanos / 1000;
+      return ms + nanos / nanosPerMs;
```
zzorba, line 172: good catch here

zzorba (approved):
LGTM

PR #84, opened 2024-08-30
Hide private fields in tagged enums and errors

### LINKED ISSUE #75: Warnings about private fields being exposed on generated types

jhugman:
According to [The Big O of Code Reviews](https://www.egorand.dev/the-big-o-of-code-reviews/), this is a O(_m * n_) change.

Potentially fixes #75.

This PR does a couple of things for TaggedEnums:

- we eliminate the `Variant__data` intermediate type and replace it with a macro to render the type explicitly. This is to make the values and their types show up in autocomplete.
- we move or remove `__uniffiTypeName`, `__variant` and `__variantName` from `UniffiEnum` and `UniffiError`. I think that this fixes the bug #75. If not, I don't think it's harmful to keep this fix.
- we introduce a `playground` directory into the `typescript/tests` directory: this is explicitly for experimenting with hand writing generated code before touching the templates.


zzorba (approved):
LGTM, this works great and eliminates the TS error I was seeing earlier

PR #85, opened 2024-09-02
Better error stack traces for async calls

### LINKED ISSUE #79: Improvement for callstacks with async function
Today, an async function generates code like this:

```
  public async getMediaContent(
    mediaSource: MediaSourceInterface
  ): Promise<ArrayBuffer> /*throws*/ {
    return await uniffiRustCallAsync(
      /*rustFutureFunc:*/ () => {
        return nativeModule().uniffi_matrix_sdk_ffi_fn_method_client_get_media_content(
          uniffiTypeClientObjectFactory.clonePointer(this),
          FfiConverterTypeMediaSource.lower(mediaSource)
        );
      },
      /*pollFunc:*/ nativeModule()
        .ffi_matrix_sdk_ffi_rust_future_poll_rust_buffer,
      /*cancelFunc:*/ nativeModule()
        .ffi_matrix_sdk_ffi_rust_future_cancel_rust_buffer,
      /*completeFunc:*/ nativeModule()
        .ffi_matrix_sdk_ffi_rust_future_complete_rust_buffer,
      /*freeFunc:*/ nativeModule()
        .ffi_matrix_sdk_ffi_rust_future_free_rust_buffer,
      /*liftFunc:*/ FfiConverterArrayBuffer.lift.bind(FfiConverterArrayBuffer),
      /*liftString:*/ FfiConverterString.lift,
      /*errorHandler:*/ FfiConverterTypeClientError.lift.bind(
        FfiConverterTypeClientError
      )
    );
  }
  ```
  
  Unfortunately, this means that if an error is returned, it will appear in the asyncDispatcher in hermes, with very little callstack present.
  
  I believe that if we changed the generated code to include a try/catch/rethrow and an await, it would at least let us locate the error to the calling async function, instead of in a random part of the underlying system.
  
  ```
  public async getMediaContent(
    mediaSource: MediaSourceInterface
  ): Promise<ArrayBuffer> /*throws*/ {
+ try {
    return await uniffiRustCallAsync(
      /*rustFutureFunc:*/ () => {
        return nativeModule().uniffi_matrix_sdk_ffi_fn_method_client_get_media_content(
          uniffiTypeClientObjectFactory.clonePointer(this),
          FfiConverterTypeMediaSource.lower(mediaSource)
        );
      },
      /*pollFunc:*/ nativeModule()
        .ffi_matrix_sdk_ffi_rust_future_poll_rust_buffer,
      /*cancelFunc:*/ nativeModule()
        .ffi_matrix_sdk_ffi_rust_future_cancel_rust_buffer,
      /*completeFunc:*/ nativeModule()
        .ffi_matrix_sdk_ffi_rust_future_complete_rust_buffer,
      /*freeFunc:*/ nativeModule()
        .ffi_matrix_sdk_ffi_rust_future_free_rust_buffer,
      /*liftFunc:*/ FfiConverterArrayBuffer.lift.bind(FfiConverterArrayBuffer),
      /*liftString:*/ FfiConverterString.lift,
      /*errorHandler:*/ FfiConverterTypeClientError.lift.bind(
        FfiConverterTypeClientError
      )
    );
  + } catch (e) {
  + throw e;
  + }
  }
  ```

This would make the errors a lot more actionable and would probably lead to a lot fewer false positives.

### PR DESCRIPTION
**Pull request by jhugman**
Potentially fixes #79.

### DIFF
--- a/typescript/src/async-rust-call.ts
+++ b/typescript/src/async-rust-call.ts
@@ -115,7 +115,9 @@ const UNIFFI_RUST_FUTURE_RESOLVER_MAP = new UniffiHandleMap<
 
 // pollRust makes a new promise, stores the resolver in the resolver map,
 // then calls the pollFunc with the handle.
-function pollRust(pollFunc: (handle: UniffiHandle) => void): Promise<number> {
+async function pollRust(
+  pollFunc: (handle: UniffiHandle) => void,
+): Promise<number> {
   return new Promise<number>((resolve) => {
     const handle = UNIFFI_RUST_FUTURE_RESOLVER_MAP.insert(resolve);
     pollFunc(handle);
--- a/crates/ubrn_bindgen/src/bindings/react_native/gen_typescript/templates/macros.ts
+++ b/crates/ubrn_bindgen/src/bindings/react_native/gen_typescript/templates/macros.ts
@@ -100,19 +100,15 @@
 
 {%- macro call_body(obj_factory, callable) %}
 {%- if callable.is_async() %}
-    {%- if config.is_debug() %}
-    const _stack = new Error().stack;
+    const __stack = uniffiIsDebug ? new Error().stack : undefined;
     try {
-    {%- endif %}
-    return {# space #}{%- call call_async(obj_factory, callable) %};
-    {%- if config.is_debug() %}
-    } catch (_error: any) {
-        if (_error instanceof Error) {
-            _error.stack = _stack;
+        return {# space #}{%- call call_async(obj_factory, callable) %};
+    } catch (__error: any) {
+        if (uniffiIsDebug && __error instanceof Error) {
+            __error.stack = __stack;
         }
-        throw _error;
+        throw __error;
     }
-    {%- endif %}
 {%- else %}
 {%-     match callable.return_type() -%}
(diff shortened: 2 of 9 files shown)

### CODE REVIEW

File: typescript/src/async-rust-call.ts [hunk 1]
```diff
@@ -63,6 +63,7 @@ export async function uniffiRustCallAsync<F, T>(
 
   // We now poll the Rust future until it's ready.
   // The poll, complete and free methods are specialized by the FFIType of the return value.
+  let _stack = new Error().stack;
```
jhugman, line 66: 🥴 This would be done on every async call.

Is generating the stack trace of the same order of magnitude of performance of the FFI?

Is it worth it?

zzorba: In a debug build, probably worth it, or at least worth trying maybe?  In a release build, probably not?

I think we could guard this with a `process.env.NODE_ENV !== 'production'` conditional

zzorba (commented):
WDYT of making it conditional based on runtime?  I agree that extracting a full stack one very call seems... not ideal

jhugman:
I've updated this and merged it with another PR: it's now keyed off a `logLevel` set in `uniffi.toml`. I'm hoping to be able to switch that on a command line switch later.

PR #87, opened 2024-09-04
Deterministic code generation order

jhugman:

This PR adds determinism to the generation of the code. This allows generated code to be checked into version control and it not be a hot mess each time a change is made.

### DIFF
--- a/crates/ubrn_bindgen/src/bindings/react_native/mod.rs
+++ b/crates/ubrn_bindgen/src/bindings/react_native/mod.rs
@@ -311,7 +311,9 @@ impl ComponentInterface {
 
         let mut sorted: Vec<Type> = Vec::new();
         while !graph.peek_all().is_empty() {
-            sorted.extend(graph.pop_all().iter().filter_map(|name| types.remove(name)));
+            let mut next = graph.pop_all();
+            next.sort();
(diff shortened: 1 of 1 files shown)


crates/ubrn_bindgen/src/bindings/react_native/mod.rs
```diff
@@ -311,7 +311,9 @@ impl ComponentInterface {
 
         let mut sorted: Vec<Type> = Vec::new();
         while !graph.peek_all().is_empty() {
-            sorted.extend(graph.pop_all().iter().filter_map(|name| types.remove(name)));
+            let mut next = graph.pop_all();
+            next.sort();
+            sorted.extend(next.iter().filter_map(|name| types.remove(name)));
```
jhugman, line 316: The topological sort has a deterministic input, sorted alphabetically.

Its output was not: each time multiple nodes had zero incoming edges, the order of those nodes was non-deterministic (it was stored in a `HashMap`).

This change sorts them with a stable sort (alphabetically).

zzorba: Very clever, this is a detail I would have totally overlooked but is super important.

zzorba (approved):
LGTM

PR #92, opened 2024-09-10
Detect if ArrayBuffer is really a typed array

### LINKED ISSUE #91: ArrayBuffer has lax enforcement in Typescript
It seems like because of historic javascript reasons, Typescript is lax about the enforcement of ArrayBuffers.  Because `Uint8Array` implements `ArrayBufferView` (i.e. it has a `.buffer` call that gives you an ArrayBuffer), it does not complain when you pass a Uin8Array to something that expects an ArrayBuffer.  This is apparently because many JS interfaces declare they take an ArrayBuffer, but actually accept both, using the following gem of code:
```  const arrayBuffer = input instanceof Uint8Array ? input.buffer : input;```

This under-documented quirk is unfortunate.

I would propose we update our `FfiConverterArrayBuffer` to explicitly take either an `ArrayBuffer | ArrayBufferView` and do the above operation.  We can update the places where ArrayBuffer is passed in to also use the union type, or rely on the lax interpreter.

```

export const FfiConverterArrayBuffer = (() => {
  const lengthConverter = FfiConverterInt32;
  class FFIConverter extends FfiConverterPrimitive<ArrayBuffer> {
    read(from: RustBuffer): ArrayBuffer {
      const length = lengthConverter.read(from);
      return from.readBytes(length);
    }
    write(value: ArrayBuffer, into: RustBuffer): void {
      const length = value.byteLength;
      lengthConverter.write(length, into);
      into.writeBytes(value);
    }
    allocationSize(value: ArrayBuffer): number {
      return lengthConverter.allocationSize(0) + value.byteLength;
    }
  }
  return new FFIConverter();
})();
```

jhugman:
Fixes #91.

Typescript accepts a typed array (`Uint32Array`, and friends) where it expects an `ArrayBuffer`.

This could mean passing an `Uint32Array` slice where a `Vec<u8>` is expected.

This commit ensures the `FfiConverterArrayBuffer` handles this correctly: it now sends the slice of the underlying buffer that the typed array represents.

Unfortunately, it is not possible to detect pathological cases:

```ts
const ab = new ArrayBuffer(256);
const ta8 = new Uint8Array(ab, 0, 64); // view of `ab`.
const ta32 = new Uint32Array(ta8, 0, 16); // nested view of view of `ab`.
````

I would like the `ffi-converter` to detect this case, but it can't.

This is because the constructor of `Uint32Array` treats the value of `ta8` as an `ArrayLike` iterable, rather than an `ArrayBufferView` or `ArrayBuffer`.

This should likely be documented.

### DIFF
--- a/fixtures/coverall2/tests/bindings/test_coverall2.ts
+++ b/fixtures/coverall2/tests/bindings/test_coverall2.ts
@@ -42,6 +42,52 @@ test("array buffer roundtrip using read/write", (t) => {
   }
 });
 
+test("array buffer roundtrip with ArrayBufferView", (t) => {
+  function rt(ab: ArrayBuffer) {
+    t.assertEqual(
+      ab,
+      identityArrayBuffer(new Uint32Array(ab)),
+      undefined,
+      abEquals,
+    );
+  }
+  for (let i = 0; i < 64; i += 4) {
+    rt(arrayBuffer(i));
+  }
+});
+
+test("array buffer roundtrip with ArrayBufferView of different sizes", (t) => {
+  function rt(ta: ArrayBuffer, slice: ArrayBuffer) {
+    t.assertEqual(slice, identityArrayBuffer(ta), undefined, abEquals);
+  }
+  const base = arrayBuffer(64);
+  for (const TypedArray of [
+    Uint8Array,
+    Uint16Array,
+    Uint32Array,
+    BigUint64Array,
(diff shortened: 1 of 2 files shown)


zzorba (approved):
LGTM

Just when you think Typescript has grown up and can sit at the big-kids table, it goes and does something like this.

PR #94, opened 2024-09-10
Mark fields as non private

zzorba:
So my previous fix did not work, my verification technique must have been flawed.  The private designation was still breaking us.

Unfortunately, it seems like Typescript does not have a way to restrict access to the fields while still letting them be accessed.  We can either tell people to ignore the errors, or just leave them exposed.  I think leaving them exposed is probably better.


jhugman:
> Unfortunately, it seems like Typescript does not have a way to restrict access to the fields while still letting them be accessed. We can either tell people to ignore the errors, or just leave them exposed.

Given how many times we've filed and fixed this bug, perhaps we've been relying on Typescript too much. Looking at the `Symbol` docs:

> Symbols are often used to add unique property keys to an object that won't collide with keys any other code might add to the object, and which are hidden from any mechanisms other code will typically use to access the object. That enables a form of weak [encapsulation](https://developer.mozilla.org/en-US/docs/Glossary/Encapsulation), or a weak form of [information hiding](https://en.wikipedia.org/wiki/Information_hiding).

This feels like the Javascript way of approaching the problem.

*update* testing out this, I got further than I thought. I'll push that into another PR.

jhugman:
Closing in favour of #95

PR #95, opened 2024-09-11
Changed __ prefixed private fields to symbols

jhugman:

This supersedes #94 to hide properties that are only used internally within the generated code.

It introduces a `symbols.ts` file and replaces all references to `__uniffiTypeName`, `__variant`, `__rustPointer` and `__rustArcPtr`.

The private accessors have also been removed.

The tests all continue to work, and the properties do not show up in the autocomplete/content assist of VS Code.


zzorba (approved):
LGTM

Issue #77, opened 2024-08-22
Building two simulator frameworks requires using lipo on the artifacts


zzorba:
If you specify both an x86 and an arm64 simulator in your .yaml file:

```
ios:
  directory: ios
  targets:
    - aarch64-apple-ios
    - aarch64-apple-ios-sim
    - x86_64-apple-ios
```

Building on iOS will attempt to create an xcframework, but it will fail because it contains TWO simulator targets.  Based on the code in the matrix-rust-sdk, https://github.com/matrix-org/matrix-rust-sdk/blob/main/xtask/src/swift.rs, the correct thing to do in this situation is to use `lipo` on the two simulator lib files to produce a single fat lib.  This can then be passed successfully to the xcframework command.

I don't know how urgent this is to FIX FIX, given that x86 is on the way out.  But having uniffi bindgen detect both x86 and arm64 simulators and call it out as unsupported seems better.

Adding an option to skip the .xcframework creation would also be a helpful escape hatch.

jhugman:
> Adding an option to skip the .xcframework creation would also be a helpful escape hatch.

For later readers, the `--no-xcodebuild` / `--no-xcframework` escape hatch was landed in https://github.com/jhugman/uniffi-bindgen-react-native/pull/71/commits/cade14de0b1ded65eb3a3f5e1ab96cb19aa12e91.

This escape hatch on its own does not fix this issue.

PR #96, opened 2024-09-12
Add lipo support to allow multiple targets of the same platform

jhugman:
According to [The Big O of Code Reviews](https://www.egorand.dev/the-big-o-of-code-reviews/), this is a O(_m *n_) change.

This PR makes two changes to the build command:

- Adds a command line override of the targets that are being built. This list of valid targets are found by looking at `ubrn build ios --help` and its android counterpart.
- Adds lipo support. This should help in the cases where multi-architecture builds are needed. (Fixes: #77)

The lipo support is largely copied from the matrix-rust-sdk build.

### DIFF
--- a/crates/ubrn_cli/src/ios.rs
+++ b/crates/ubrn_cli/src/ios.rs
@@ -4,13 +4,13 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/
  */
 
-use std::{collections::HashMap, process::Command};
+use std::{collections::HashMap, fmt::Display, process::Command, str::FromStr};
 
-use anyhow::Result;
+use anyhow::{Context, Error, Result};
 use camino::{Utf8Path, Utf8PathBuf};
 use clap::Args;
 use heck::ToUpperCamelCase;
-use serde::Deserialize;
+use serde::{Deserialize, Serialize};
 use ubrn_common::{mk_dir, rm_dir, run_cmd, CrateMetadata};
 
 use crate::{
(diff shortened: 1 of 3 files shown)


crates/ubrn_cli/src/ios.rs
```diff
@@ -190,10 +208,44 @@ impl IOsArgs {
...
+            let files = by_platform.entry(target.platform).or_insert(Vec::new());
+            files.push(file);
+        }
+        let metadata = crate_.metadata()?;
+
+        let mut sorted = Vec::new();
+        for (p, mut files) in by_platform {
+            if files.len() == 1 {
+                sorted.append(&mut files);
+            } else {
```
zzorba, line 227: Is lipo needed on android?  We haven't had to use it so far, but I guess they don't have separate 'simulator' targets?

crates/ubrn_cli/src/ios.rs
```diff
@@ -241,3 +297,97 @@ impl IOsArgs {
...
+}
+
+impl Platform {
+    /// The name of the subfolder in which to place the library for the platform
+    /// once all architectures are lipo'd together.
+    fn lib_folder_name(&self) -> &str {
+        match self {
+            Platform::Macos => "macos",
+            Platform::Ios => "ios",
+            Platform::IosSimulator => "ios-simulator",
```
zzorba, line 325: Some part of me thinks that XCode might be very particular about the names of these folders, but I don't have a citation to back that up?

zzorba (approved):
LGTM

crates/ubrn_cli/src/ios.rs
jhugman: > Is lipo needed on android?

No, `lipo` is part of the Apple toolchain.

crates/ubrn_cli/src/ios.rs
jhugman: The way I'm using them at the moment is to create a folder in the Rust `target` directory, as outputs of the lipo command.

The `xcodebuild -create-xcframework` then uses those files to create the `Framework.xcframework` folder.

PR #98, opened 2024-09-26
Tighten up logic around generated codegened spec filename

jhugman:

Fixes #44.

This PR investigates and fixes how the React Native Codegen file is named, taking into consideration the rules around what it creates.

It turns out that it:

 * has to start with `Native`
 * it is the file name that is used to generate all the other files 
 * the spec name in pacakge.json seems to be irrelevant.

The PR still allows ReactNative and RN to be trimmed, but only if it is not specified in the configuration YAML.

E.g. `name: RnDiode` to the top of the `config.yaml`.


zzorba (approved):
LGTM

PR #100, opened 2024-09-30
Further re-work `trim_react_native`

jhugman:
Reverts #98
Closes #44.

#98 did not seem to fix them problem, so reverting, and cherry-picking https://github.com/jhugman/uniffi-bindgen-react-native/pull/100/commits/47878b1282f8def961da5ae71053949d02c3fb47 .

There are two extra things in #98: ordering the modules in a consistent manner so that the generated turbo-modules can be checked in to VCS, and returning `true` in the `installRustCrate` method.


zzorba (approved):
LGTM

PR #102, opened 2024-09-30
Translates cli into javascript

jhugman:
This PR translates the command line launcher in to JS.

This means it can be launched from `package.json`.

Adding to `package.json`:

```json
"scripts": {
    "ubrn": "uniffi-bindgen-react-native"
}
```

means now we can do:

```sh
yarn ubrn --help
```

As a little extra:

```sh
alias ubrn=$(yarn ubrn --path)
```
will make the command invokable from the shell.


zzorba (approved):
very nice

PR #103, opened 2024-09-30
Allow record default properties to be overridden

### LINKED ISSUE #101: Generated create method for structs makes it hard to use default values.
When looking at a uniffi record like this, it generates a helper class that contains create/default methods. 

Rust code:
```
#[derive(uniffi::Record)]
pub struct CreateRoomParameters {
    pub name: Option<String>,
    #[uniffi(default = None)]
    pub topic: Option<String>,
    pub is_encrypted: bool,
    #[uniffi(default = false)]
    pub is_direct: bool,
    pub visibility: RoomVisibility,
    pub preset: RoomPreset,
    #[uniffi(default = None)]
    pub invite: Option<Vec<String>>,
    #[uniffi(default = None)]
    pub avatar: Option<String>,
    #[uniffi(default = None)]
    pub power_level_content_override: Option<PowerLevels>,
}
```

Typescript code:
```
export type CreateRoomParameters = {
  name: string | undefined;
  topic: string | undefined;
  isEncrypted: boolean;
  isDirect: boolean;
  visibility: RoomVisibility;
  preset: RoomPreset;
  invite: Array<string> | undefined;
  avatar: string | undefined;
  powerLevelContentOverride: PowerLevels | undefined;
};

/**
 * Generated factory for {@link CreateRoomParameters} record objects.
 */
export const CreateRoomParameters = (() => {
  const defaults = () => ({
    topic: undefined,
    isDirect: false,
    invite: undefined,
    avatar: undefined,
    powerLevelContentOverride: undefined,
  });
  const create = (() => {
    return uniffiCreateRecord<
      CreateRoomParameters,
      ReturnType<typeof defaults>
    >(defaults);
  })();
  return Object.freeze({
    /**
     * Create a frozen instance of {@link CreateRoomParameters}, with defaults specified
     * in Rust, in the {@link matrix_sdk_ffi} crate.
     */
    create,

    /**
     * Create a frozen instance of {@link CreateRoomParameters}, with defaults specified
     * in Rust, in the {@link matrix_sdk_ffi} crate.
     */
    new: create,

    /**
     * Defaults specified in the {@link matrix_sdk_ffi} crate.
     */
    defaults: () => Object.freeze(defaults()) as Partial<CreateRoomParameters>,
  });
})();
```

I think the problem stems from the uniffiCreateRecord method:
```
export const uniffiCreateRecord = <T, D extends Partial<T>>(
  defaults: () => D,
) => {
  type MissingKeys = Omit<T, keyof D>;
  return (partial: Required<MissingKeys> & Partial<D>): T =>
    Object.freeze({ ...defaults(), ...partial } as T);
};
```

Because defaults has a return type that contains only the default values (in this example, undefined and false), the use of Partial<D> here artificially restricts the allowed values.

I think the correct fix here would be to use Pick to grab the full types:

```
type MissingKeys = Omit<T, keyof D>;
type OptionalKeys = Pick<T, keyof D>;
return (partial: Required<MissingKeys> & Partial<OptionalKeys>): T =>
    Object.freeze({ ...defaults(), ...partial } as T);
```

jhugman:
Fixes #101 .

This PR adds a test for the behaviour described in #101, and a fix.

### DIFF
--- a/typescript/src/records.ts
+++ b/typescript/src/records.ts
@@ -12,6 +12,6 @@ export const uniffiCreateRecord = <T, D extends Partial<T>>(
   defaults: () => D,
 ) => {
   type MissingKeys = Omit<T, keyof D>;
-  return (partial: Required<MissingKeys> & Partial<D>): T =>
+  return (partial: Partial<T> & Required<MissingKeys>): T =>
     Object.freeze({ ...defaults(), ...partial } as T);
 };
--- a/typescript/tests/playground/records.ts
+++ b/typescript/tests/playground/records.ts
@@ -1,0 +1,30 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/
+ */
+
+import { uniffiCreateRecord } from "../../src/records";
+
+export type MyRecord = {
+  string: string;
+  number: number;
+  optionalString: string | undefined;
+  bool: boolean;
+  optionalBool: boolean | undefined;
+};
+
+export const MyRecord = (() => {
+  const defaults = () => ({
+    string: "default",
+    optionalString: undefined,
+  });
+
+  const create = uniffiCreateRecord<MyRecord, ReturnType<typeof defaults>>(
+    defaults,
+  );
+
+  return {
+    create,
+  };
+})();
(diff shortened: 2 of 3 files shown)


typescript/src/records.ts
```diff
@@ -12,6 +12,6 @@ export const uniffiCreateRecord = <T, D extends Partial<T>>(
   defaults: () => D,
 ) => {
   type MissingKeys = Omit<T, keyof D>;
-  return (partial: Required<MissingKeys> & Partial<D>): T =>
+  return (partial: Partial<T> & Required<MissingKeys>): T =>
```
jhugman, line 15: This is the fix! The types were the wrong way round.

zzorba: Ah-ha!

zzorba (approved):
LGTM

PR #108, opened 2024-10-02
Try CJS extension to make script runnable

zzorba:


jhugman (approved):
While linking to `./bin/cli` is working for me, I'm hoping that linking to `./bin/cli.cjs` works for both of us.

kraenhansen:
The file referenced by the `bin` property of a package.json, must be executable, meaning, it must start with the "node shebang". Or as the docs put it: https://docs.npmjs.com/cli/v10/configuring-npm/package-json#bin

> Please make sure that your file(s) referenced in bin starts with #!/usr/bin/env node, otherwise the scripts are started without the node executable!

kraenhansen:
@zzorba I'm curious what issues you saw before this change?

zzorba:
It was a while ago, but I believe I was getting complaints about it not being a runnable target.

It quite likely this was not the correct fix, I'm open to trying something else.

kraenhansen:
> not being a runnable target.

This can happen if the file is not marked as executable - I've found that I have to set the execute permission on files like this in the past. Luckily git and the .tar.gz uploaded to NPM remembers this, so it just have to happen once:

```
chmod u+x ./bin/cli
```

zzorba:
I'll test this out tomorrow.

zzorba:
Just to close the loop, change is working great for me locally.

Issue #109, opened 2024-10-02
Generated code is being prefixed (e.g. with "ReactNative") but some occurrences appear to be missing


Johennes:
I have the following in `package.json`:

```json5
"codegenConfig": {
  "name": "RNMatrixSdkSpec",
  "type": "modules",
  "jsSrcsDir": "src"
}
```

When generating the turbo module with `46f81224e4aae9a05dc7c8e209b4750c014f4d02` (main branch from an hour ago), the iOS module comes out as

```
// Generated by uniffi-bindgen-react-native
#ifdef __cplusplus
#import "react-native-matrix-sdk.h"
#endif

#ifdef RCT_NEW_ARCH_ENABLED
#import "RNMatrixSdkSpec.h"

@interface ReactNativeMatrixSdk : NSObject <RNMatrixSdkSpec>
#else
#import <React/RCTBridgeModule.h>

@interface ReactNativeMatrixSdk : NSObject <RCTBridgeModule>
#endif

@end
```

Notice how it imports `RNMatrixSdkSpec.h` and tries to implement the eponymous protocol `<RNMatrixSdkSpec>`. However, the generated file in `example/ios/build/generated/ios/RNMatrixSdkSpec/RNMatrixSdkSpec.h` defines the protocol as `@protocol NativeReactNativeMatrixSdkSpec`. As a result, the build on iOS fails.

When I change the name in `package.json` to `NativeReactNativeMatrixSdkSpec`

```json5
"codegenConfig": {
  "name": "NativeReactNativeMatrixSdkSpec",
  "type": "modules",
  "jsSrcsDir": "src"
}
```

both the generated file and the protocol come out as `NativeReactNativeMatrixSdkSpec` and the build succeeds.

Johennes:
Possibly related: The Android build fails, too, after switching to `46f81224e4aae9a05dc7c8e209b4750c014f4d02`. The problem there is that my `android/src/main/AndroidManifest.xml` used `com.matrixsdk` as the package name (which is what `react-native-create-library` had generated for my `react-native-matrix-sdk` repository).

```
<manifest xmlns:android="http://schemas.android.com/apk/res/android"
          package="com.matrixsdk">
</manifest>
```

The generated code now uses `com.reactnativematrixsdk`, however.

Changing the package name in `AndroidManifest.xml` fixes thing. Maybe UBRN should generate this file, too?

jhugman:
Thanks for your analysis!

[...]

PR #72, opened 2024-08-17
Draft documentation

jhugman:
First cut of the reference sections of the documentation.

The https://github.com/jhugman/uniffi-bindgen-react-native/pull/72/commits/48a01cac80e0ba400bc66fd5b625b8df5daf0195 commit is likely the most interesting to readers and reviewers.

This is not ready to merge, but should be instructive nevertheless.

This is likely best reviewed via `mdbook serve docs`.

PR #112, opened 2024-10-04
Auto-generate AndroidManifest.xml

Johennes:
Relates to: #109

I'm not sure if this is strictly necessary anymore because once I had it done, I realized that #72 has stripped the `reactnative` prefix from the package name again. So technically UBRN's current output should work with what create-react-native-library generates without this change applied. It might not hurt either way though?


jhugman (approved):
Perfect!

PR #115, opened 2024-10-07
Check out shallowly and allow updating a checkout

### LINKED ISSUE #113: Allow checking out SHA
UBRN currently allows checking out a specific branch or tag via the `--branch` flag. It's probably a small use case but it might be nice if UBRN could also check out a specific commit based on its SHA. This has the advantage that it cannot easily be swapped out unnoticed by the upstream repository.

### LINKED ISSUE #114: Updating a checkout
When changing the branch _after_ the initial checkout, UBRN will fail when reattempting to check out because the repository already exists. The workaround is deleting `rust_modules` and rerunning the checkout command which, depending on the repository, can take quite some time.

It'd be great if UBRN could simply update the existing cloned repository.

As a further performance optimization, maybe it could use a shallow clone and then only fetch the specified reference on update? For matrix-rust-sdk, a shallow clone takes ~2s for me vs. ~20 for the full clone.

Johennes:
This fixes #114 by:

* Performing a shallow clone if the directory doesn't already exist
* Using a combination of `git fetch` and `git checkout` to shallowly check out the supplied branch

This unintentionally also fixes #113 because the git commands used work for both refs and SHAs.

### DIFF
--- a/docs/src/api/config-yaml.md
+++ b/docs/src/api/config-yaml.md
@@ -28,7 +28,7 @@ rust:
 	branch: main
 	manifest-path: crates/my-api/Cargo.toml
 ```
-In this case, the `ubrn checkout` command will clone the given repo with the branch/ref into the `rust_modules` directory of the project.
+In this case, the `ubrn checkout` command will clone the given repo with the branch/ref into the `rust_modules` directory of the project. Note that instead of `branch` you can also use `rev` or `ref`.
 
 If run a second time, no overwriting will occur.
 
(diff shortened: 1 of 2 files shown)


jhugman (approved):
This looks lovely.

Should we consider aliasing the `branch` to `ref` or `rev` (or both)?

```rust
#[serde(alias = "rev", alias = "ref", default = "GitRepoArgs::default_branch")]
```

docs/src/api/config-yaml.md
```diff
@@ -28,7 +28,7 @@ rust:
 	branch: main
 	manifest-path: crates/my-api/Cargo.toml
 ```
-In this case, the `ubrn checkout` command will clone the given repo with the branch/ref into the `rust_modules` directory of the project.
+In this case, the `ubrn checkout` command will clone the given repo with the branch/ref into the `rust_modules` directory of the project. Note that instead of `branch` you can also use `rev` or `ref`.
```
jhugman, line 31: 👨‍🍳 🤌 

Chef's kiss. Thank you so much.

PR #116, opened 2024-10-08
Documentation part II

jhugman:
This PR fleshes out the documentation book.

It should be fairly readable in Markdown in the github interface, but may be better done in mdbook.

Note to self: document how to contribute to documents.


zzorba (approved):
LGTM

PR #117, opened 2024-10-09
Add tag property to flat errors

### LINKED ISSUE #111: Improvement for uniffi flat errors
In the uniffi exposed library we are using, it seems like the use of the flat_error definition is fairly common:

```
#[derive(Debug, thiserror::Error, uniffi::Error)]
#[uniffi(flat_error)]
pub enum RegisterAccountError {
    #[error("Password is too weak")]
    WeakPassword,
    #[error("Email already in use")]
    EmailAlreadyInUser,
    #[error("Verification needed")]
    VerificationNeeded,
    #[error("An error occurred: {msg}")]
    Generic { msg: String },
}
```

Today to unpack these, it seems like you have to use the individual instanceOf methods:
```
catch (err) {
  if (RegisterAccountError.WeakPassword.instanceOf(err))  {
     // process
   } else if (RegisterAccountError.EmailAlreadyInUse.instanceOf(err)) {
      //proces
    } // etc
  }
  ```
  
It would be nice if there was some kind of generated typescript enum that could be matched against in Typescript for these, to help ensure you are exhaustive in your type checking.  I would imagine it was like the tag that is used for other Enums in the system, something like this:
```
catch (err) {
   if (RegisterAccountError.instanceOf(err)) {
     switch(err.tag) {
       case RegisterAccountErrorTag.WeakPassword: 
          // stuff
         break;
      case RegisterAccountErrorTag.EmailAlreadyInUse:
        // more stuff
        break;
     default:
    }
 }
```

I know Errors can be a bit of an 'anything' in rust, which makes it hard to bridge to typescript.  But I was thinking maybe the flat_error marker could be something we key off of here?

jhugman:
Fixes #111.

This PR adds a `tag` property and a `MyError_Tags` enum to flat errors.

to the error template, it adds some testing of the template, and some docs.

### DIFF
--- a/crates/ubrn_bindgen/src/bindings/react_native/gen_typescript/templates/ErrorTemplate.ts
+++ b/crates/ubrn_bindgen/src/bindings/react_native/gen_typescript/templates/ErrorTemplate.ts
@@ -1,10 +1,18 @@
 {{- self.import_infra("UniffiError", "errors") }}
 {{- self.import_infra("uniffiTypeNameSymbol", "symbols") }}
 {{- self.import_infra("variantOrdinalSymbol", "symbols") }}
-{%- let instance_of = "instanceOf" %}
 {%- let flat = e.is_flat() %}
 {%- if flat %}
+
 // Flat error type: {{ decl_type_name }}
+{%- let type_name__Tags = format!("{type_name}_Tags") %}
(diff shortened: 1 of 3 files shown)


zzorba (approved):
LGTM

PR #123, opened 2024-10-11
Trim react-native from the name in the same places create-rn-library and Codegen does

### LINKED ISSUE #109: Generated code is being prefixed (e.g. with "ReactNative") but some occurrences appear to be missing

jhugman:
Fixes #109.

This PR adds testing around the turbo-modules.

The tests runs through a host of different names, then 

- runs `create-react-native-library` with that name. 
- runs the `ubrn generate turbo-module` command
- the checks that generated files from the ubrn command still uses the identifiers in the `create-react-native-library` files
- finally builds for ios and android the library.

This has meant a series of changes (for the better) to the templates.


jhugman:
I haven't yet got the test running in CI, but I think that can wait for another PR. Marking ready for review.

crates/ubrn_cli/src/codegen/mod.rs
```diff
@@ -264,7 +264,7 @@ mod files {
     templated_file!(IndexTs, "index.ts");
```
Johennes, line 264: Should this also end in `.tsx`?

jhugman: Yes, you're right. I renamed the template file and made this `index.tsx` to match. I also renamed the `IndexTs` struct.

Thanks for prompting me out of laziness. :)

zzorba (approved):
LGTM

Issue #126, opened 2024-10-15
ubrn checkout gives `detached HEAD` warning to stderr


jhugman:
When running ubrn checkout, even for a default branch of `main`, a `detached HEAD` warning is echoed to stderr.

e.g. running 

```sh
./scripts/run-checkout-tests.sh 2>&1
```

Observed:

```sh
[TEST] checkout with default
Running "git" "clone" "https://github.com/actions/checkout" "/Users/jhugman/workspaces/react-native-uniffi/work/uniffi-bindgen-react-native/rust_modules/checkout" "--depth" "1"
Cloning into '/Users/jhugman/workspaces/react-native-uniffi/work/uniffi-bindgen-react-native/rust_modules/checkout'...
Running cd "/Users/jhugman/workspaces/react-native-uniffi/work/uniffi-bindgen-react-native/rust_modules/checkout" && "git" "fetch" "--depth" "1" "origin" "eef61447b9ff4aafe5dcd4e0bbf5d482be7e7871"
From https://github.com/actions/checkout
 * branch            eef61447b9ff4aafe5dcd4e0bbf5d482be7e7871 -> FETCH_HEAD
Running cd "/Users/jhugman/workspaces/react-native-uniffi/work/uniffi-bindgen-react-native/rust_modules/checkout" && "git" "checkout" "eef61447b9ff4aafe5dcd4e0bbf5d482be7e7871"
Note: switching to 'eef61447b9ff4aafe5dcd4e0bbf5d482be7e7871'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by switching back to a branch.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -c with the switch command. Example:

  git switch -c <new-branch-name>

Or undo this operation with:

  git switch -

Turn off this advice by setting config variable advice.detachedHead to false

HEAD is now at eef6144 Prepare 4.2.1 release (#1925)
~/workspaces/react-native-uniffi/work/uniffi-bindgen-react-native/rust_modules/checkout ~/workspaces/react-native-uniffi/work/uniffi-bindgen-react-native
✅ OK
~/workspaces/react-native-uniffi/work/uniffi-bindgen-react-native
[TEST] checkout with branch
[...]

PR #127, opened 2024-10-15
Allow turbo-modules with names with @my-org/project-name

### LINKED ISSUE #124: Name mangling breaks for iOS in organization-scoped packages
In an organization scoped-package such as `@unomed/react-native-matrix-sdk`, UBRN puts the podspec into `./@unomed/react-native-matrix-sdk.podspec`. However, the file actually needs to go into the root folder as `unomed-react-native-matrix-sdk.podspec`.

Similarly, inside the file, UBRN uses `s.name = "@unomed/react-native-matrix-sdk"` while it should be `s.name  = "unomed-react-native-matrix-sdk"`.

When I move the file, change the name and rebase the paths, everything works fine again.

PS: I also tested Android in an organization-scoped package. Everything appears to work fine out of the box there. 👌

jhugman:
Fixes #124 

This adds to the tests introduced in #123 to support names with an org name.

### DIFF
--- a/crates/ubrn_cli/src/codegen/mod.rs
+++ b/crates/ubrn_cli/src/codegen/mod.rs
@@ -327,7 +327,7 @@ mod files {
     templated_file!(PodspecTemplate, "module-template.podspec");
     impl RenderedFile for PodspecTemplate {
         fn path(&self, project_root: &Utf8Path) -> Utf8PathBuf {
-            let name = self.config.project.raw_name();
+            let name = self.config.project.cpp_filename();
             let filename = format!("{name}.podspec");
             project_root.join(filename)
         }
--- a/crates/ubrn_cli/src/codegen/mod.rs
+++ b/crates/ubrn_cli/src/codegen/mod.rs
@@ -327,7 +327,7 @@ mod files {
     templated_file!(PodspecTemplate, "module-template.podspec");
     impl RenderedFile for PodspecTemplate {
         fn path(&self, project_root: &Utf8Path) -> Utf8PathBuf {
-            let name = self.config.project.cpp_filename();
+            let name = self.config.project.podspec_filename();
             let filename = format!("{name}.podspec");
             project_root.join(filename)
         }
--- a/scripts/test-turbo-modules.sh
+++ b/scripts/test-turbo-modules.sh
@@ -374,50 +374,50 @@ run_default() {
   local config="$fixture_dir/ubrn.config.yaml"
   local app_tsx="$fixture_dir/App.tsx"
   main \
-        --force-new-directory \
-        --keep-directory-on-exit \
-        --ubrn-config "$config" \
-        --builder-bob-version 0.35.1 \
-        --skip-ios \
-        --skip-android \
-        --slug dummy-lib \
-        "$working_dir/dummy-lib"
+    --force-new-directory \
+    --keep-directory-on-exit \
+    --ubrn-config "$config" \
+    --builder-bob-version 0.35.1 \
+    --skip-ios \
+    --skip-android \
+    --slug dummy-lib \
+    "$working_dir/dummy-lib"
   main \
-        --force-new-directory \
-        --keep-directory-on-exit \
-        --ubrn-config "$config" \
-        --builder-bob-version 0.35.1 \
-        --skip-ios \
-        --skip-android \
-        --slug rn-dummy-lib \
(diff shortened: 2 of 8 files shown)


scripts/run-checkout-tests.sh
```diff
@@ -40,6 +40,7 @@ assert_eq $(git ls-remote --heads origin | grep refs/heads/releases/v1 | cut -f1
 popd
 
 clean_up
+exit 0
```
Johennes, line 43: I think this might be a leftover?

scripts/test-turbo-modules.sh
```diff
@@ -401,6 +402,51 @@ run_default() {
...
   # ReactNativeDummyLib fails with "› Must be a valid npm package name"
+  main \
+    --force-new-directory \
+    --keep-directory-on-exit \
+    --ubrn-config "$config" \
+    --builder-bob-version 0.35.1 \
+    --skip-ios \
+    --skip-android \
+    --slug @my-org/react-native-dummy-lib \
+    "$working_dir/@my-org/react-native-dummy-lib"
```
Johennes, line 413: The lines below use different indentation. Not that it matters at all of course. 😅

```suggestion
      --force-new-directory \
      --keep-directory-on-exit \
      --ubrn-config "$config" \
      --builder-bob-version 0.35.1 \
      --skip-ios \
      --skip-android \
      --slug @my-org/react-native-dummy-lib \
      "$working_dir/@my-org/react-native-dummy-lib"
```

crates/ubrn_cli/src/codegen/mod.rs
```diff
@@ -327,7 +327,7 @@ mod files {
     templated_file!(PodspecTemplate, "module-template.podspec");
     impl RenderedFile for PodspecTemplate {
         fn path(&self, project_root: &Utf8Path) -> Utf8PathBuf {
-            let name = self.config.project.raw_name();
+            let name = self.config.project.cpp_filename();
```
Johennes, line 330: It strikes me as slightly odd that the filename for C++ is reused for the podspec. It doesn't really make a huge difference if the value is the same but maybe using a dedicated function would be cleaner?

Johennes (commented):
I tested this on https://github.com/unomed-dev/react-native-matrix-sdk using the example app and things work fine for iOS. On Android, the app appears to crash. At least I think it does. When I tap the button to make the native call in the app, it looks as if it goes into the background unexpectedly but then seems to relaunch entirely when I try to foreground it. I haven't seen this behaviour before and I'm not getting _any_ related output on the terminal. 🤔

zzorba:
I'll test out this build on my side @Johennes to see if I get similar results.  I don't have a namespaced package like this, but I'm guessing the crash you are seeing is unrelated to that?

Johennes:
> I'll test out this build on my side @Johennes to see if I get similar results. I don't have a namespaced package like this, but I'm guessing the crash you are seeing is unrelated to that?

Possibly though I'm not entirely sure since the organisation ends up in some generated names but not others. I'll try to also give it some more testing tomorrow.

Johennes:
Yeah, I noticed that, too, but already had `index` without any extension in `package.json` so didn't have to change anything. I think I'll try running the app from Android Studio later today to see that gives me an actual error message or stack trace.

scripts/run-checkout-tests.sh
jhugman: Yes! That was trying to get the error messages for #126 !

Removed now.

jhugman:
> > I'll test out this build on my side @Johennes to see if I get similar results. I don't have a namespaced package like this, but I'm guessing the crash you are seeing is unrelated to that?
> 
> Possibly though I'm not entirely sure since the organisation ends up in some generated names but not others. I'll try to also give it some more testing tomorrow.

I found it!

The Android crash was due to the library name in `System.loadLibrary` not matching the one built in `CMakeLists.txt`. 

I think I broke it in #123, but might have broken it in this one.

I've changed this and added a test.

jhugman:
> One thing I noticed here, it seems like it went from `src/index.ts` to `src/index.tsx`, so I had to update my package.json slightly.

Our new tests check for differences between the generated files. Builder bob was generating an `index.tsx` file, and `ubrn` was generating an `index.ts` file.

zzorba:
Confirmed the fix is working!

zzorba (approved):
LGTM

PR #128, opened 2024-10-15
Silence git checkout warnings

Johennes:
This silences the `git checkout` warnings. I wasn't sure if you wanted to remove output from the other git commands, too?

Fixes: #126

### DIFF
--- a/crates/ubrn_cli/src/repo.rs
+++ b/crates/ubrn_cli/src/repo.rs
@@ -111,10 +111,11 @@ impl GitRepoArgs {
             .arg(sha);
         run_cmd(&mut cmd)?;
 
-        // git checkout $sha
+        // git checkout --quiet $sha
         let mut cmd = Command::new("git");
         cmd.current_dir(self.directory(project_root)?)
             .arg("checkout")
+            .arg("--quiet") // Suppress detached head and dangling commit warnings
             .arg(sha);
         run_cmd(&mut cmd)
     }
--- a/crates/ubrn_cli/src/repo.rs
+++ b/crates/ubrn_cli/src/repo.rs
@@ -79,21 +79,33 @@ impl GitRepoArgs {
             run_cmd(&mut cmd)?;
         }
 
-        // git ls-remote origin $branch
+        // git remote set-branches origin '*' to start tracking all branches and
+        // enable checking out branch names
+        let mut cmd = Command::new("git");
+        cmd.current_dir(self.directory(project_root)?)
+            .arg("remote")
+            .arg("set-branches")
+            .arg("origin")
+            .arg("*");
+        run_cmd(&mut cmd)?;
+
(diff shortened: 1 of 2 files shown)


crates/ubrn_cli/src/repo.rs
```diff
@@ -111,10 +111,11 @@ impl GitRepoArgs {
             .arg(sha);
         run_cmd(&mut cmd)?;
 
-        // git checkout $sha
+        // git checkout --quiet $sha
         let mut cmd = Command::new("git");
         cmd.current_dir(self.directory(project_root)?)
             .arg("checkout")
+            .arg("--quiet") // Suppress detached head and dangling commit warnings
```
jhugman, line 118: I don't _love_ this, but it does solve the proximal problem.

I think (and it doesn't have to be in this PR) I would prefer to not get into a detached HEAD state in the first place, not just to keep quiet about it.

When you're checking out to a tag, it'll be impossible to avoid getting into a detached HEAD state and we should warn, but when it's a branch then we should avoid it.

Unless you want to keep going on it, perhaps we should land this now and file another issue to enable this more elaborate fix. What do you think?

jhugman (approved):

crates/ubrn_cli/src/repo.rs
Johennes: I kind of feel obliged to fix the mess I created and sorry for causing it in the first place. 😕

I pushed another attempt. This calls `git remote set-branches origin '*'` after cloning which enables tracking of all remote branches in the shallow fork. That in turn enables checking out branch names without getting into a detached HEAD state. I also added tests to assert which checkout targets cause detached HEAD states:

* Checkout without target: no detached HEAD
* Checkout with branch name: no detached HEAD
* Checkout with tag name: detached HEAD
* Checkout with SHA: detached HEAD

This appears to work. Alas, the checkout logic has become somewhat involved overall. If you're not happy with this, maybe it'll be better to revert back to non-shallow clones. I do enjoy the speediness of shallow checkouts but I suppose most people won't be doing checkouts overly often.

Johennes: I've colored the test headings in the script output so you can more easily see the output of each checkout command in https://github.com/jhugman/uniffi-bindgen-react-native/actions/runs/11371743080/job/31634502135?pr=128 now.

jhugman: >  If you're not happy with this, maybe it'll be better to revert back to non-shallow clones.

I am very happy. Thank you so much.

>  I do enjoy the speediness of shallow checkouts but I suppose most people won't be doing checkouts overly often. So maybe I've been over-optimizing here.

Hard agree. Let's see how people use this, and then optimize for that use case if it ever needs to be optimized.

> I've colored the test headings in the script output 

Love it. It's the little things that add that feel of polish.

PR #129, opened 2024-10-15
Split CI into several jobs

Johennes:
I noticed that the CI job takes quite long but most of the time is spent in the code generation tests. This splits up the one job into several which gives better progress visualization and faster status feedback on PRs.

I guess I got carried away a little with the emoji. 😅 They didn't turn out as fancy as I thought though. Maybe you have better ideas or otherwise I can just remove them again.

While at it, I also added a status badge in the `README`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ env:
 
 jobs:
   lint:
-    name: Lint
+    name: 🔍 Lint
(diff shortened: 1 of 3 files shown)


zzorba (approved):
Thank you for this, very good

jhugman:
Oh, A M A Z I N G. No notes. Merged without hesitation.

jhugman (approved):

Issue #121, opened 2024-10-10
Document steps for how to publish a uniffi'd project on npmjs


Johennes:
After setting up UBRN in a new project created with create-react-native-library@0.35.1, I had to add the following to `.gitignore` to exclude the Rust checkouts and binaries from git:

```
# uniffi-bindgen-react-native
rust_modules/
*.a
```

When packing the library for publishing, npm appears to have special logic for combining `.gitignore` and the `files` array from `package.json`.

For Android, `*.a` in `.gitignore` and `"android"` in `files` results in the `android/src/main/jniLibs/$target/$library.a` files being included in the package.

For iOS, however, in order to include `build/$library.xcframework`, I had to add `"build"` to `files` to undo the `build/` from `.gitignore`.

I'm not yet sure how to best publish packages built with UBRN but decided to log this here for documentation in the meanwhile. Changes to `.gitignore` and or `files` appear to be needed regardless of whether one wants to include the binaries in the package or not.

jhugman:
In truth, I haven't spent very much time writing Javascript/Typescript, or in this community.

I don't really know how people are going to want to distribute uniffied packages via npm.

I am kinda waiting for—/ hoping that— the community (including you!) will work out what to do next.

At a minimum, I can add some `.gitignore` instructions to the tutorial, but I'd love to see more patterns of usage and more developer input before writing either code or more comprehensive docs.

I'm going to add the label:help+wanted tag to this issue.

Johennes:
I've successfully published https://www.npmjs.com/package/@unomed/react-native-matrix-sdk with the binaries for all targets except for x86 on the public npm registry yesterday. The unpacked size is ~430MB. Packed is about a third.

Annoyingly, I wasn't able to find any official reference for the maximum allowed package or file size. So I'm not sure how much head room there is.

Including the binaries in the package has the huge advantage that consumers don't need to spend an hour building matrix-rust-sdk in release mode.

PR #130, opened 2024-10-16
Expand publishing docs

Johennes:
Fixes: #121

I should add that my own understanding of the npm publishing process isn't super extensive. So this will probably benefit from improvement in the future.

One thing I thought about when writing this up: If we could get UBRN to output the `.xcframework` into the `/ios` folder, the default `.gitignore` and `package.json` that `create-react-native-library` generates might actually be sufficient without any modifications.

### DIFF
--- a/docs/src/guides/publishing.md
+++ b/docs/src/guides/publishing.md
@@ -8,17 +8,26 @@ I would love some [help with this document](../contributing/documentation.md).
 
 ## Binary builds
 
-In order to distribute pre-built packages you will need to add to `.gitignore` the built `.a`:
+You likely don't want to track pre-built binaries in your git repository but you may want to include them in published packages. If so, you will have to work around `npm`'s behaviour of factoring in `.gitignore` when picking files to include in packages.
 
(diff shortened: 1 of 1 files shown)


docs/src/guides/publishing.md
```diff
@@ -35,6 +44,8 @@ scripts: {
 +   "postinstall":   "yarn ubrn:checkout && yarn ubrn:android && yarn ubrn:ios",
```
jhugman, line 44: I think this:

```suggestion
+   "postinstall":   "yarn ubrn:checkout && yarn ubrn:android --release && yarn ubrn:ios --release",
```

will obviate the need to your new paragraph.

jhugman (approved):
One small nit, but otherwise, lovely. Thank you.

docs/src/guides/publishing.md
Johennes: Ah, great!

jhugman:
Thanks @Johennes !

PR #131, opened 2024-10-17
create-react-native-library@latest: Modify turbo-modules generation

### LINKED ISSUE #107: Doesn't integrate well with create-react-native-library > 0.35.1
I believe people using UBRN are likely to depend on create-react-native-library to bootstrap a library and then integrate UBRN into it. With UBRN `4adf6d9190ea2eac852e3be7ee83b0b2271e4b18` and create-react-native-library@0.35.1 this works pretty flawlessly. The only thing required is adding `pod 'uniffi-bindgen-react-native', :path => '../../node_modules/uniffi-bindgen-react-native'` in `example/iosPodfile`.

Things are much more tricky with the currently latest version of create-react-native-library, 0.41.2, though. It lets you pick between two different ways to set up the example app, "test app" and "vanilla".

The "test app" variant works well on Android but not on iOS. It's using `use_test_app!` in the `Podfile` and lets you add the UBRN pod like this:

```
use_test_app! do |target|
  target.app do
    pod 'uniffi-bindgen-react-native', :path => '../../node_modules/uniffi-bindgen-react-native'
  end
end
```

When running the app, however, any call into the turbo module throws "error in host function" claiming the native implementation is missing.

The "vanilla" variant left me with the opposite problem. The example app ran on iOS but not on Android.

I didn't dig deeper but my impression was that further code generation is triggered during the build of the example app but with different output paths than UBRN's code generation.

jhugman:
Fixes #107

Now working with `create-react-native-library@0.41.2`

This adds kotlin detection and uses Kotlin if needed, or stays with Java: this adds two Kotlin files and an alternative `build.gradle` file.

It also adds more checks to the test-turbo-modules.sh script, tightening up podspec and Objective C files.

Finally we add documentation to getting-started.md and contributing-turbo-modules.md.


jhugman:
@Johennes may also be interested in this PR.

Johennes:
Thanks for taking this one. Looks like a lot of work. 💪

I tested this by regenerating @unomed/react-native-matrix-sdk with create-react-native-library@latest. The example app works for me on iOS but not on Android:

```
/Users/jm/Code/react-native-matrix-sdk/android/src/main/java/com/unomed/reactnativematrixsdk/ReactNativeMatrixSdkModule.java:11: error: cannot find symbol
public class ReactNativeMatrixSdkModule extends NativeReactNativeMatrixSdkSpec {
```

I do have the `NativeReactNativeMatrixSdkSpec` class output from RN's codegen though.

```
$ cat android/generated/java/com/unomed/reactnativematrixsdk/NativeReactNativeMatrixSdkSpec.java

/**
 * This code was generated by [react-native-codegen](https://www.npmjs.com/package/react-native-codegen).
 *
 * Do not edit this file as changes may cause incorrect behavior and will be lost
 * once the code is regenerated.
 *
 * @generated by codegen project: GenerateModuleJavaSpec.js
 *
 * @nolint
 */

package com.unomed.reactnativematrixsdk;

import com.facebook.proguard.annotations.DoNotStrip;
import com.facebook.react.bridge.ReactApplicationContext;
import com.facebook.react.bridge.ReactContextBaseJavaModule;
import com.facebook.react.bridge.ReactMethod;
import com.facebook.react.turbomodule.core.interfaces.TurboModule;
import javax.annotation.Nonnull;

public abstract class NativeReactNativeMatrixSdkSpec extends ReactContextBaseJavaModule implements TurboModule {
  ...
```

I didn't have a lot of time to dig into it tonight sadly.

jhugman:
Hmm. This is annoying. 

Could you run this from within your checked out ubrn directory, substituting the variables:

```sh
./scripts/test-turbo-modules.sh \
  --ubrn-config $YOUR_UBRN_CONFIG_FILE \
  --app-tsx $YOUR_APP_TSX_FILE \
  --builder-bob-version latest \
  --android \
  --ios \
  --keep-directory-on-exit \
  --force-new-directory \
  --slug @unomed/react-native-matrix-sdk \
  --ios-name ReactNativeMatrixSdk \
  "$DIRECTORY"
```

I am a bit confused why you're seeing java files in `src/main/java/com/unomed/reactnativematrixsdk`; I would expect you to see Kotlin.

Johennes:
> I am a bit confused why you're seeing java files in src/main/java/com/unomed/reactnativematrixsdk; I would expect you to see Kotlin.

Funnily I had the exact same thought overnight and dug into that this morning. What happened is I wanted to `rm -r android/build` before building the example app because I'm frequently getting build failures when switching between ubrn versions because of some cached build artifacts. However, I fat-fingered and deleted the entire `android` folder instead. That killed `build.gradle` which you're using to determine whether to generate Kotlin or Java.

I copied `build.gradle` over from a project freshly generated with create-react-native-library, reran the ubrn generation et voilà the Android build works. 🎉

So the issue was entirely on my part, apologies. 🤦‍♂️ 

It does feel slightly circular to read the Kotlin flag from a file that ubrn itself generates. I couldn't think of another way to do it either though and if people don't mess up the create-react-native-library generation like I did, things should be fine.

I also tested ubrn from this branch on the repo that I had generated with the old create-react-native-library version and things worked fine there on both platforms.

Thanks again for making ubrn work with the latest create-react-native-library!

jhugman:
> It does feel slightly circular to read the Kotlin flag from a file that ubrn itself generates. I couldn't think of another way to do it either though and if people don't mess up the create-react-native-library generation like I did, things should be fine.

I did consider checking in package.json for the creat-react-nativw-library version but couldn't see when they'd started adding versions. Also version number sniffing felt too much like useragent sniffing.

Thanks so much for re-testing. That is such a relief.

jhugman:
Could you review this please @zzorba

zzorba (approved):
LGTM, I tested this locally and my android build is still working just fine (and I think it uses kotlin).

Issue #134, opened 2024-10-18
default config doesn't build for relevant architecture on Intel Mac


trevoranderson:
The default config is

```
---
name: MyRustLib
rust:
  repo: https://github.com/ianthetechie/uniffi-starter
  branch: main
  manifestPath: rust/foobar/Cargo.toml
```

When it builds on iOS it only builds Arm targetes, but on an Intel Mac, the simulator is x86, so it fails to link.
The following config builds on Intel Mac:

```
---
name: MyRustLib
rust:
  repo: https://github.com/ianthetechie/uniffi-starter
  branch: main
  manifestPath: rust/foobar/Cargo.toml
ios:
    directory: ios
    targets:
    - aarch64-apple-ios
    - aarch64-apple-ios-sim
    - x86_64-apple-ios
```

PR #135, opened 2024-10-19
Default to x86_64-apple-ios if ubrn is built for x86

Johennes:
This adds `x86_64-apple-ios` rather than `aarch64-apple-ios-sim` to the default targets on iOS if UBRN itself was built for x86.

Given that x86 Macs are slowly becoming extinct, I think, it may not be worth to include both targets by default because it'll result in a lot of pointless compilation.

I believe the `ARCH` value is baked in at compile time. So this may not work correctly in some edge cases. @trevoranderson, maybe you could give this a test on your Intel Mac?

Fixes: #134

### DIFF
--- a/crates/ubrn_cli/src/ios.rs
+++ b/crates/ubrn_cli/src/ios.rs
@@ -4,7 +4,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/
  */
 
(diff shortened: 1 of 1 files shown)


zzorba (approved):
this makes a lot of sense IMO.

LGTM

jhugman:
This is ingenious; I had not thought of this solution. I'm going to land it now.

PR #143, opened 2024-10-24
Partial support for duplicated symbols

### LINKED ISSUE #68: Handle namespaced type references from other files/packages
Does UBRN handle namespaced names?  Not if they are used in the same file (package?)

Theoretically I could have code that refers to two different rust packages of the same name, from different packages, for example both `matrix_sdk::client::Client` and `matrix_sdk_ffi::client::Client` would have the short name of Client. Or I could have a local class and an imported class referred to in the same package, one with a qualified name.

Sample rust code:

```
#[derive(uniffi::Enum, Clone, Debug)]
pub enum ClientBuilder {
    Foo,
    Bar,
}

#[derive(Clone, uniffi::Enum)]
pub enum WrappingEnum {
    TotallyFine,
    FirstClient(ClientBuilder),
    SecondClient(Arc<matrix_sdk_ffi::client_builder::ClientBuilder>),
}
```

Thankfully, this doesn't produce incorrect code today, but it does produce an assertion in UBRN and the generate process fails as a result.

jhugman:

This PR partially fixes #68:

- it re-exports each named crate as default export.
- it documents working with megazords.
- including the limitations of the Swift backend.

The second part to fix #68 is a little more involved, but is not yet worth it:

- for imported types and their converters, the bindings should use a more fully qualified name. e.g.

Instead of:

```ts
import { Crate1 } from "./crate1";
import crate1 from "./crate1";

const { FfiConverterTypeCrate1 } = crate1.converters;

// In use
type MyType = {
  scalar: Crate1,
  optional: Crate1 | undefined,
  list: Array<Crate1>,
}

// FfiConverterTypeCrate1 being used.
const lift = FfiConverterTypeCrate1.lift.bind(FfiConverterTypeCrate1);
```

The next part needs to generate code:

```ts
import * as crate1 from "./crate1";

// In use
type MyType = {
  scalar: crate1.Crate1,
  optional: crate1.Crate1 | undefined,
  list: Array<crate1.Crate1>,
}

// FfiConverterTypeCrate1 being used.
const lift = crate1.converters.FfiConverterTypeCrate1.lift.bind(crate1.converters.FfiConverterTypeCrate1);
```

I think I will file a new issue with this in, and close #68.

### DIFF
--- a/docs/src/guides/megazords.md
+++ b/docs/src/guides/megazords.md
@@ -39,7 +39,7 @@ In this case, your library's client code would import `Crate1Type` and `Crate2Ty
 import { Crate1Type, Crate2Type } from "react-native-my-megazord";
 ```
 
-Alternatively, they can the default export:
+Alternatively, they can use the default export:
 
 ```ts
 import megazord from "react-native-my-megazord";
--- a/docs/src/guides/megazords.md
+++ b/docs/src/guides/megazords.md
@@ -1,0 +1,55 @@
+# Working with multiple crates in one library
+
+Some teams arrange their Rust library in to multiple crates, or multiple teams from one organization combine their efforts into one library.
+
+This might be for better code organization, or to reduce shipping multiple copies of the same dependencies.
+
+The combined library from multiple crates, in Mozilla vernacular, is known as a [Megazord](https://robots.fandom.com/wiki/Mighty_Morphin%27_Megazord).
+
+`uniffi-rs` and `uniffi-bindgen-react-native` both work well with Megazords.
+
+`uniffi-bindgen-react-native` produces a cluster of files per crate. For example, generating files from the library `libmymegazord.a` might contain two crates, `crate1` and `crate2`. The library directory would look like this:
+
+```
+cpp
+├── generated
+│   ├── crate1.cpp
+│   ├── crate1.hpp
+│   ├── crate2.cpp
+│   └── crate2.hpp
+├── react-native-my-megazord.cpp
+└── react-native-my-megazord.h
+src
+├── NativeMyMegazord.ts
+├── generated
+│   ├── crate1.ts
+│   ├── crate1-ffi.ts
+│   ├── crate2.ts
+│   └── crate2-ffi.ts
+└── index.tsx
+```
+
+In `index.tsx`, the types are re-exported from `crate1.ts` and `crate2.ts`.
+
+In this extended example, `crate1.ts` might declare a `Crate1Type` and `crate2.ts` a `Crate2Type`.
+
+In this case, your library's client code would import `Crate1Type` and `Crate2Type` like this:
+
+```ts
+import { Crate1Type, Crate2Type } from "react-native-my-megazord";
+```
+
+Alternatively, they can the default export:
+
+```ts
+import megazord from "react-native-my-megazord";
+
+const { Crate1Type } = megazord.crate1;
+const { Crate2Type } = megazord.crate2;
+```
+
+```admonish warning title="Duplicated identifiers"
+Due to Swift's large granular module sytem, crates in the same megazord cannot have types of the same name.
+
+This may be solved in Swift at some point— e.g. by adding prefixes— but until then, duplicate identifiers will cause a Typescript compilation error as the types are smooshed together in `index.tsx`.
+```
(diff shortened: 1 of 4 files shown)


docs/src/guides/megazords.md
```diff
@@ -0,0 +1,55 @@
...
+
+In this extended example, `crate1.ts` might declare a `Crate1Type` and `crate2.ts` a `Crate2Type`.
+
+In this case, your library's client code would import `Crate1Type` and `Crate2Type` like this:
+
+```ts
+import { Crate1Type, Crate2Type } from "react-native-my-megazord";
+```
+
+Alternatively, they can the default export:
```
Johennes, line 42:
```suggestion
Alternatively, they can use the default export:
```

docs/src/guides/megazords.md
```diff
@@ -0,0 +1,55 @@
...
+import megazord from "react-native-my-megazord";
+
+const { Crate1Type } = megazord.crate1;
+const { Crate2Type } = megazord.crate2;
+```
+
+```admonish warning title="Duplicated identifiers"
+Due to Swift's large granular module sytem, crates in the same megazord cannot have types of the same name.
+
+This may be solved in Swift at some point— e.g. by adding prefixes— but until then, duplicate identifiers will cause a Typescript compilation error as the types are smooshed together in `index.tsx`.
```
Johennes, line 54:
```suggestion
This may be solved in Swift at some point — e.g. by adding prefixes — but until then, duplicate identifiers will cause a Typescript compilation error as the types are smooshed together in `index.tsx`.
```

Johennes (approved):
Reviewed and tested with https://github.com/unomed-dev/react-native-matrix-sdk and I think this looks good. Here are the changes it caused for reference:

```
diff --git a/src/index.tsx b/src/index.tsx
index 242ba02..36133d2 100644
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -12,17 +12,31 @@ export * from './generated/matrix_sdk_crypto';
 export * from './generated/matrix_sdk_ffi';
 export * from './generated/matrix_sdk_ui';
 
+// Now import the bindings so we can:
+// - intialize them
+// - export them as namespaced objects as the default export.
+import * as matrix_sdk from './generated/matrix_sdk';
+import * as matrix_sdk_base from './generated/matrix_sdk_base';
+import * as matrix_sdk_common from './generated/matrix_sdk_common';
+import * as matrix_sdk_crypto from './generated/matrix_sdk_crypto';
+import * as matrix_sdk_ffi from './generated/matrix_sdk_ffi';
+import * as matrix_sdk_ui from './generated/matrix_sdk_ui';
+
 // Initialize the generated bindings: mostly checksums, but also callbacks.
-import matrix_sdk_ from './generated/matrix_sdk';
-import matrix_sdk_base_ from './generated/matrix_sdk_base';
-import matrix_sdk_common_ from './generated/matrix_sdk_common';
-import matrix_sdk_crypto_ from './generated/matrix_sdk_crypto';
-import matrix_sdk_ffi_ from './generated/matrix_sdk_ffi';
-import matrix_sdk_ui_ from './generated/matrix_sdk_ui';
+matrix_sdk.default.initialize();
+matrix_sdk_base.default.initialize();
+matrix_sdk_common.default.initialize();
+matrix_sdk_crypto.default.initialize();
+matrix_sdk_ffi.default.initialize();
+matrix_sdk_ui.default.initialize();
+
+// Export the crates as individually namespaced objects.
+export default {
+  matrix_sdk,
+  matrix_sdk_base,
+  matrix_sdk_common,
+  matrix_sdk_crypto,
+  matrix_sdk_ffi,
+  matrix_sdk_ui,
+};
 
-matrix_sdk_.initialize();
-matrix_sdk_base_.initialize();
-matrix_sdk_common_.initialize();
-matrix_sdk_crypto_.initialize();
-matrix_sdk_ffi_.initialize();
-matrix_sdk_ui_.initialize();
```

docs/src/guides/megazords.md
Johennes: Oh ok. I wasn't aware it's a particular style, apologies. If that's the case, we can just keep it like this everywhere.

jhugman:
Swift work seems to being tracked in [this issue](https://github.com/mozilla/uniffi-rs/issues/2220).

PR #145, opened 2024-10-25
Support abortController.abort() call after the future has finished

### LINKED ISSUE #144: AbortController causes crash if called after the future has completed
A common use case for `AbortController` is to cancel async operations that were started via the `useEffect` hook.

```typescript
useEffect(() => {
  const controller = new AbortController();

  async function loadLatestEvent() {
    if (!item.isTimelineInitialized()) {
      await item.initTimeline(undefined, undefined, { signal: controller.signal });
    }
    ...
  }

  loadLatestEvent().catch(...);

  return () => {
    controller.abort();
  }
}, [item]);
```

This appears to cause a crash when the clean-up callback runs _after_ the promise has completed.

```
Thread 2 Crashed:: com.facebook.react.runtime.JavaScript
0   unomed_mobile.debug.dylib     	       0x10f28c9d0 ffi_matrix_sdk_base_rust_future_cancel_f32 + 44
```

The code generated by UBRN should be able to recognise this situation and simply do nothing if the abort signal arrives after the task completed.

jhugman:
Fixes #144.

If an `AbortSignal` is passed to a Promise returning function, and the promise returns then this should do nothing.

This commit unregister the AbortSignal callback after the Promise has been resolved or rejected.

This does not need any additional documentation.

### DIFF
--- a/fixtures/futures/tests/bindings/test_futures.ts
+++ b/fixtures/futures/tests/bindings/test_futures.ts
@@ -497,6 +497,42 @@ function checkRemainingFutures(t: Asserts) {
   );
 
   await asyncTest(
+    "future method… which is cancelled after it is resolved",
+    async (t) => {
+      const abortController = new AbortController();
+      const expected = await fallibleMe(false, {
+        signal: abortController.signal,
+      });
+      t.assertEqual(expected, 42);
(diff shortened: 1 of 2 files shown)


zzorba (approved):
LGTM

PR #151, opened 2024-10-31
Documentation for option/result and use of undefined vs null

zzorba:
Please feel free to revise as needed.


docs/src/idioms/option-result.md
```diff
@@ -0,0 +1,12 @@
+# Options and Nullables
+
+In keeping with uniffi guidelines, we have made an effort to map core Rust concepts into their Typescript equivalents wherever possible, in order to allow the Typescript code to be as idiomatic and easy to work with as possible. A returned `Result<T, E>` from Rust will result in a flat type of `T` or a thrown `E`, while an `Option<T>` will become `T | undefined`. We believed that flattened types like this would be easier to work with than wrappers at every point for such common Rust primitives. This in in keeping with the approach taken by the core uniffi team for other languages (Kotlin turns `Option<T>` into `T`?).
```
jhugman, line 3:
```suggestion
In keeping with uniffi guidelines, we have made an effort to map core Rust concepts into their Typescript equivalents wherever possible, in order to allow the Typescript code to be as idiomatic and easy to work with as possible. A returned `Result<T, E>` from Rust will result in a flat type of `T` or a thrown `E`, while an `Option<T>` will become `T | undefined`. We believed that flattened types like this would be easier to work with than wrappers at every point for such common Rust primitives. This is in keeping with the approach taken by the core uniffi team for other languages (Kotlin turns `Option<T>` into `T`?).
```

docs/src/idioms/option-result.md
```diff
@@ -0,0 +1,12 @@
...
+
+Note that this flattening does mean certain types which are perfectly legal (if ill advised) in Rust are not representable on the Uniffi layer. An `Option<Option<T>>` for example, could be represented in Rust, though we would assert it may be a poor stylistic choice even there. If you truly need to represent a tri-state, a wrapping enum with data feels like a clearer choice, and one that has first class support. We also note that the core uniffi project shares these limitations in some of its bindings (Kotlin), and it has not proven overly burdensome to date.
```
jhugman, line 5:
```suggestion
Note that this flattening does mean certain types which are perfectly legal (if ill advised) in Rust are not representable on the Uniffi layer. An `Option<Option<T>>` for example, could be represented in Rust, though we would assert it may be a poor stylistic choice even there. If you need to represent a tri-state, an enum with three variants feels like a clearer choice, and one that has first class support. We also note that the core uniffi project shares these limitations in some of its bindings (Kotlin), and it has not proven overly burdensome to date.
```

jhugman (approved):
Great! I had no idea you were going to do this. Thank you!

zzorba:
Yeah, had some time on the subway this morning so was a good 20 minutes activity.

PR #152, opened 2024-10-31
Generate nested optionals avoiding degenerate case

jhugman:
`Option<Option<T>>` used to generate `T | undefined | undefined`. This PR removes the second and subsequent `| undefined`s.


zzorba (approved):
LGTM

PR #153, opened 2024-11-03
Fix missing cmakelist libs for android with RN 0.76

ronickg:
When using with RN 0.76 fixes the build error for missing linked libs. Only have tested on android, but shouldn't be a problem for ios.

```
npx create-react-native-library@latest my-rust-lib
```

Both "ReactAndroid::turbomodulejsijni" and "ReactAndroid::react_nativemodule_core" have been removed.


ronickg:
Oh my bad just saw there already is an open pull request.

PR #154, opened 2024-11-06
Fix the configuration errors in the document about ubrn.config.yaml

kingsword09:
There's a problem with the [android](https://jhugman.github.io/uniffi-bindgen-react-native/reference/config-yaml.html#android) configuration in the `ubrn.config.yaml` file in the document.

When I configure android according to the document, the following error occurs:
```bash
# Error: Failed to read "ubrn.config.yaml" as valid YAML

# Caused by:
#     android.targets[0]: unknown variant `aarch64-linux-android`, expected one of `armeabi-v7a`, `arm64-v8a`, `x86`, `x86_64` at line 10 column 5
```
After looking at the [code](https://github.com/jhugman/uniffi-bindgen-react-native/blob/main/crates/ubrn_cli/src/android.rs#L135), I found that it does need to be changed to Abi `armeabi-v7a`, `arm64-v8a`, `x86`, `x86_64`.

### DIFF
--- a/docs/src/reference/config-yaml.md
+++ b/docs/src/reference/config-yaml.md
@@ -70,10 +70,10 @@ android:
 	directory: ./android
 	cargoExtras: []
 	targets:
-	- aarch64-linux-android
-	- armv7-linux-androideabi
-	- i686-linux-android
-	- x86_64-linux-android
+	- arm64-v8a
+	- armeabi-v7a
+	- x86
+	- x86_64
 	apiLevel: 21
 	jniLibs: src/main/jniLibs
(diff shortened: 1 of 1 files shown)


Johennes (approved):
This looks good and evidently matches what e.g. https://github.com/unomed-dev/react-native-matrix-sdk/blob/main/ubrn.yaml#L10 uses.

jhugman:
Brilliant! Thanks @kingsword09 .

I'll investigate further why this was failing (it shouldn't) but will merge this nevertheless. Thank you.

Issue #150, opened 2024-10-30
Switch from askama to rinja


jhugman:
Following https://github.com/mozilla/uniffi-rs/pull/2222 , for consistency this project should move to [rinja](https://rinja.readthedocs.io/en/stable/) also.

jhugman:
This is a pre-requisite of bumping uniffi-rs version up from 0.28.

PR #155, opened 2024-11-06
Migrate from askama to rinja

Johennes:
Fixes: #150

I started working on this before realizing it'll first require a new release of uniffi-rs. So leaving this here as a draft to be picked up again later.


crates/ubrn_bindgen/src/bindings/react_native/gen_typescript/templates/macros.ts
```diff
@@ -237,19 +237,19 @@
     {%- endif %}
 {%- endmacro %}
 
-{%- macro async(func) %}
+{%- macro is_async(func) %}
```
Johennes, line 240: This renaming and the ones below are required because rinja doesn't allow Rust keywords as macro names. I'm not sure if the names are particularly great but this follows what uniffi-rs itself does.

jhugman:
> I started working on this before realizing it'll first require a new release of uniffi-rs. So leaving this here as a draft to be picked up again later.

I don't think this relies on uniffi-rs, but there maybe some more `self::` and `borrow`s to add to the templates. Would @GuillaumeGomez be able to help here?

Johennes:
I haven't dug deeply but this error looked as if it was still pulling in askama (even though its not a direct dependency anymore) and the only crate that appears to do it was from uniffi-rs.

```
115 | #[derive(Template)]
    |                 ^ the trait `From<askama::error::Error>` is not implemented for `bindings::_::rinja::Error`, which is required by `Result<(), bindings::_::rinja::Error>: FromResidual<Result<Infallible, askama::error::Error>>`
```

Maybe that error is just a surrogate though?

GuillaumeGomez:
That's strange. Unless you still have `askama` in your dependency tree, you should never see `askama` anywhere as `rinja` doesn't depend on it (directly or indirectly).

Johennes:
Poking at it a little, the build fails on a number of statements in the templates, for instance:

```
{%- let obj = ci|get_object_definition(name) %}
```

If I'm reading things correctly, that function bubbles up into `lookup_error` which results in `askama::Error` in the versions of the uniffi-rs crates that we're using and was [changed](https://github.com/mozilla/uniffi-rs/pull/2222/files#diff-d9e36f4c9d8ecd97f0e2d669f86603aea311e9be99107ef197340870011f9155L34) to `rinja::Error` in the merged yet unreleased pull request that migrated uniffi-rs to rinja.

GuillaumeGomez:
Ah then until you use the latest version, I suppose you'll have the issue.

Johennes:
Closing in favor of #193.

PR #156, opened 2024-11-09
Prevent further prompts in npx create-react-native-library

Johennes:
These changes should make it easier to run `test-turbo-modules.sh` on the CI in future.

### DIFF
--- a/scripts/test-turbo-modules.sh
+++ b/scripts/test-turbo-modules.sh
@@ -165,7 +165,7 @@ create_library() {
     example_type=vanilla
   fi
   echo "-- Creating library $PROJECT_SLUG with create-react-native-library@$BOB_VERSION"
-  npx "create-react-native-library@$BOB_VERSION" \
+  npm_config_yes=true npx "create-react-native-library@$BOB_VERSION" \
     --slug "$PROJECT_SLUG" \
     --description "An automated test" \
     --author-name "James" \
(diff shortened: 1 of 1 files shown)


scripts/test-turbo-modules.sh
```diff
@@ -165,7 +165,7 @@ create_library() {
     example_type=vanilla
   fi
   echo "-- Creating library $PROJECT_SLUG with create-react-native-library@$BOB_VERSION"
-  npx "create-react-native-library@$BOB_VERSION" \
+  npm_config_yes=true npx "create-react-native-library@$BOB_VERSION" \
```
Johennes, line 168: This confirms the install prompt when you haven't run the supplied version of create-react-native-library before.

```
Need to install the following packages:
create-react-native-library@0.43.0
Ok to proceed? (y)
```

scripts/test-turbo-modules.sh
```diff
@@ -175,6 +175,7 @@ create_library() {
     --languages cpp \
     --type module-new \
     --example $example_type \
+    --local false \
```
Johennes, line 178: This declines generating a local library when running the script from the root folder of another project (such as ubrn itself).

```
? Looks like you're under a project folder. Do you want to create a local library? › (Y/n)
```

jhugman (approved):
These changes seem sensible and reasonable. I have tested them on my local machine, and they work.

Great work, thank you once again @Johennes !

✅ :shipit:

PR #158, opened 2024-11-11
Fixup callbacks deadlock when calling callback very frequently

### LINKED ISSUE #157: Foreign trait / callbacks deadlock
@tayrevor in reports in Matrix:

> I'm having an issue where it seems like calling a method too quickly on a foreign trait hangs indefinitely for older (slower) iOS devices. Essentially, I have a heavy computation that is streaming its best result via the foreign trait so the UI can visualize the computation. At first, the values come in very quickly. On newer phones, this isn't a problem but on older phones the thread hangs at the point where it calls the foreign trait. Inspecting with the XCode debugger after a pause reveals where the thread is stuck. Typically this occurs around 10-30 function calls on my test iPhone 10s with similar symptoms on friend's iPhone 13. Adding a 100ms sleep after the foreign function call to space them out works around the issue

> This is in a thread I created with `spawn(||{})` so it should already be an OS thread. It's not async in the tokio sense as I'm not using its runtime or the async keyword

> The deadlocked thread is the rust background thread that is calling the foreign trait method. There's nothing special about it

jhugman:
Fixes #157


trevoranderson:
This seems like the issue I had. I experienced this both in release builds (XCode archive and run on devices from TestFlight) and locally with a debug build run directly on my iPhone 10s test device.

For reference, this is the stack trace of the hung thread for me.
image

PR #160, opened 2024-11-11
Relax timing of setTimeout in abort controller test

### LINKED ISSUE #159: AbortController test intermittent failure.
Several PRs are now failing:

```
-- Starting the hermes event loop
❌ AbortController is polyfilled as expected Error: Duration out of range: 400 <= 399 <= 600
JS Exception: Error: Test 'AbortController is polyfilled as expected' timed out
    at call (native)
```

This test should either be changed e.g. tweaking the numbers, or removed i.e. removing the measurement altogether.

jhugman:
Fixes #159

### DIFF
--- a/typescript/tests/abort-controller.test.ts
+++ b/typescript/tests/abort-controller.test.ts
@@ -41,7 +41,7 @@ test("AbortController exists", (t) => {
 
     // we don't care too much about the accuracy of the time here, just that the
     // classes exist and work as intended.
-    await t.asyncMeasure(() => cancellableDelay(500), 500, 100);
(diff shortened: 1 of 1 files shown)


Johennes:
Looks ok. Hopefully it'll improve the situation. 👍

It's somewhat interesting that this only started flaking now.

Johennes (approved):

jhugman:
> It's somewhat interesting that this only started flaking now.

Yes, I'm also interested in this. 

The amount of our code needed to run that particular test is essentially just the test-harness, so I'm not too concerned—unless different timing issues start to crop up.

PR #162, opened 2024-11-12
Allow specifying the RN version in test-turbo-module.sh

Johennes:


jhugman (approved):
This looks good! What are you going to do with it?

jhugman:
> I wanted to see if I can set up a small test matrix that runs the script on the CI for different RN and builder-bob versions so that we can get some visibility into which combinations currently work and be notified when one of them breaks. Right now I'm basically running into the various externally triggered breakages by coincidence.
> 
> Do you think that would be helpful?

Getting this to run on CI would be amazing, if nothing else to get that area of the code tested in a more systematic way.

Having an early warning system for breakage based on the shifting sands of RN and CRNL would be super cool.

Would you be triggering this to run every time CRNL makes a release? or daily or something even cleverer?

I do worry that if CRNL breaks our build, that would stop us landing code elsewhere in the project—so I'd likely suggest that these tests trigger if a PR makes changes in the `crates/ubrn_cli` directory, or the `test-turbo-modules.sh` file itself.

Johennes:
I was thinking about nightly for a start just to get a sense of stability and then go from there? Definitely agree that we should limit its impact on PRs given that external factors could break it.

Issue #161, opened 2024-11-12
Doesn't work anymore with create-react-native-library@0.41.2 & react-native@0.76.1


Johennes:
The `codegenConfig` block now seems to include output dirs for the generated code.

```json5
  "codegenConfig": {
    "name": "RNReactNativeMatrixSdkSpec",
    "type": "all",
    "jsSrcsDir": "src",
    "outputDir": {
      "ios": "ios/generated",
      "android": "android/generated"
    },
    "android": {
      "javaPackageName": "com.unomed.reactnativematrixsdk"
    },
    "includesGeneratedCode": true
  },
```

UBRN doesn't appear to use these for Android and instead tries to include the generated code from `android/build/generated/source/codegen/java` in `android/build.gradle`.

```
          java.srcDirs += [
            // This is needed to build Kotlin project with NewArch enabled
            "${project.buildDir}/generated/source/codegen/java"
          ]
```

I tried changing the output dir for Android in `package.json` to comply with the value UBRN expects but this gave me another error. The only thing I found working so far is to manually patch `build.gradle`.

```
          java.srcDirs += [
            // This is needed to build Kotlin project with NewArch enabled
-            "${project.buildDir}/generated/source/codegen/java"
+            "${project.buildDir}/../generated/java"
          ]
```

UBRN should read the values from `package.json` and use them accordingly.

jhugman:
@Johennes I'm not sure if you want to fix this yourself, or not. If you don't then, no worries, but it will take a week or two for me to get to it. If you do, great! This is what I'd suggest:

Reading the `outputDir` from `package.json` should be a case of adding a struct to [`npm.rs`](https://github.com/jhugman/uniffi-bindgen-react-native/blob/main/crates/ubrn_cli/src/config/npm.rs#L67-L84). I think that file has some #[serde(default ="…")]` functions.

You may need to add something to `AndroidConfig` and `IosConfig` so it's accessible from the templates. The [`javaPackageName` property](https://github.com/jhugman/uniffi-bindgen-react-native/blob/bd3fb0f1904f4f8dcee940dcf6e72c7c07395317/crates/ubrn_cli/src/android.rs#L51-L53) is a reasonable template for how all this ties together.

[...]

PR #165, opened 2024-11-18
Handle non-standard codegen output dirs

Johennes:
Fixes: #161

<del>This includes the changes from https://github.com/jhugman/uniffi-bindgen-react-native/pull/163 but is currently based against `main` so that the workflows run.</del>

This merges into #163 which should then become green and can be merged to `main`.

### DIFF
--- a/crates/ubrn_cli/src/config/npm.rs
+++ b/crates/ubrn_cli/src/config/npm.rs
@@ -48,19 +48,11 @@ impl PackageJson {
     }
 
     pub(crate) fn android_codegen_output_dir(&self) -> String {
-        self.codegen_config
-            .output_dir
-            .android
-            .clone()
-            .unwrap_or("android/generated".to_string())
+        self.codegen_config.output_dir.android.clone()
     }
 
     pub(crate) fn ios_codegen_output_dir(&self) -> String {
-        self.codegen_config
-            .output_dir
-            .ios
-            .clone()
-            .unwrap_or("ios/generated".to_string())
+        self.codegen_config.output_dir.ios.clone()
     }
 
     pub(crate) fn repo(&self) -> &PackageJsonRepo {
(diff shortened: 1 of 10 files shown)


Johennes:
This currently still fails for Android.

```
   CMake Error at /home/runner/work/uniffi-bindgen-react-native/turbo-module/example/android/app/build/generated/autolinking/src/main/jni/Android-autolinking.cmake:9 (add_subdirectory):
    add_subdirectory given source
    "/home/runner/work/uniffi-bindgen-react-native/turbo-module/android/generated/jni/"
    which is not an existing directory.
```

I haven't yet found out where the `android/generated/jni` is coming from.

integration/fixtures/compat/react-native.config.js
Johennes, line 1: It's possible that UBRN should generate this file but it has a lot of options and I don't know if we'd deny any plausible uses by auto-generating it. Curious to hear other thoughts on this.

jhugman: I think it's worth it if and only if those options aren't going to change. However, I don't think we're in a position to tell just yet. 

I'm in favour of:

 * make a note in the docs that this path should match … whatever path it's meant to.
 * wait for a little while, and see what other projects do with it, particular other users of ubrn (if they turn up).
 * if people aren't making changes to it, then consider generating the file for them.

Johennes: This sounds like a good plan. 👍

I have added a note to the docs.

jhugman (approved):
Really nice, thank you. 

I have left a couple of nits, which you can ignore or not.

Also: the checks don't seem to have run on this PR. Is that something I missed?

crates/ubrn_cli/src/config/npm.rs
```diff
@@ -47,6 +47,22 @@ impl PackageJson {
...
+            .clone()
+            .unwrap_or("android/generated".to_string())
+    }
+
+    pub(crate) fn ios_codegen_output_dir(&self) -> String {
+        self.codegen_config
+            .output_dir
+            .ios
+            .clone()
+            .unwrap_or("ios/generated".to_string())
```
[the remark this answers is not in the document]
Johennes: I _think_ I've addressed this now.

Johennes:
> Also: the checks don't seem to have run on this PR. Is that something I missed?

Yes, you're right. They only run when merging to `main` but this PR is set up to merge into `johannes/compatibility-matrix`.

Let me actually merge this PR into the other one and we can continue the discussion there.

PR #168, opened 2024-12-02
Fixup tsc build breakage in test_coverall2.ts

### LINKED ISSUE #167: test_coverall2.ts(120,49): error TS2554: Expected 0-1 arguments, but got 3.
This has just started failing, and is blocking https://github.com/jhugman/uniffi-bindgen-react-native/pull/163 from landing.

I haven't investigated yet.

jhugman:
Fixes #167 

The build was breaking due to a compilation error in the Typescript test, due to a change in Typescript versions. 

You can see the code in [this Typescript playground](https://www.typescriptlang.org/play/?target=7&ts=5.6.3#code/FDAEoMwVwOwYwC4EsD2NQCcEAoEEMAuUAQQwzwE8AhKCCAUwwBpQBnAGyTnqNPOtoMMASlABvMKAC+kuGlYJQAIzyt6oALygY9AO4kylGnUbYAbABZhAbkkQUGUNjkwFoACoUADvQAmfSlAUCFAAbUlwAFUkGAQADgCKJgjQaNiARjNE5PAomIQAZgAmbJSqJABzNIRLUtzQAElYhMMklKaETLrcjuLu8HKKjtrWnNyAMXYUPEKS0ZTJ6ZqLboBdUQl6lzddJF8EAAtND28-RIA6KgBNdwBRAGUAfQAFW4AlR9uAGVuAWVuAHLuWz1eyObDseiKSEwCqHY67fYHaygGFwo4AHmUqno5yUFAQ9C+9FhhxRaPhAGotIjDhsUuAwU4GblIYpghA1IotAAGEH1XIcrmgSmokno0BYlRqPEEonisks8BCqEirQUg4s+kC3LbRQIU7+VrHHT6Tw+I38bDS+gsFUIFga0AAelAtIONiVoD1bE43GONvOHC49Gw9rtdGFoo1np14CwuEN2V9IdjApk9Qz0mAQA): updating to Typescript v5.7.2 introduces the compile time error. 

The fix presented here is the simplest dumbest fix: we no longer are relying on enumerating the type constructors, we now enumerate transitional objects that we provide, with hand written calls into a constructor.

### DIFF
--- a/fixtures/coverall2/tests/bindings/test_coverall2.ts
+++ b/fixtures/coverall2/tests/bindings/test_coverall2.ts
@@ -94,32 +94,97 @@ test("array buffer roundtrip with ArrayBufferView", (t) => {
 });
 
 test("array buffer roundtrip with ArrayBufferView of different sizes", (t) => {
-  function rt(ta: ArrayBuffer, slice: ArrayBuffer) {
-    t.assertEqual(slice, identityArrayBuffer(ta), undefined, abEquals);
+  function rt(viewName: string, ta: ArrayBuffer, slice: ArrayBuffer) {
+    t.assertEqual(
+      slice,
+      identityArrayBuffer(ta),
+      `${viewName} didn't match`,
+      abEquals,
+    );
   }
   const base = arrayBuffer(64);
-  for (const TypedArray of [
-    Uint8Array,
-    Uint16Array,
-    Uint32Array,
-    BigUint64Array,
-    Int8Array,
-    Int16Array,
-    Int32Array,
-    BigInt64Array,
-    Float32Array,
-    Float64Array,
-  ]) {
-    const width = TypedArray.BYTES_PER_ELEMENT;
-    for (let length = width; length < base.byteLength; length += width) {
+  const arrayTypes = [
+    {
+      name: "Uint8Array",
+      bytesPerElement: Uint8Array.BYTES_PER_ELEMENT,
+      new: (ab: ArrayBuffer, byteOffset: number, length: number) =>
+        new Uint8Array(ab, byteOffset, length),
+    },
+    {
+      name: "Uint16Array",
+      bytesPerElement: Uint16Array.BYTES_PER_ELEMENT,
+      new: (ab: ArrayBuffer, byteOffset: number, length: number) =>
+        new Uint16Array(ab, byteOffset, length),
+    },
+    {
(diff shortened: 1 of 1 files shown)


Johennes:
It looks like TS 5.7.2 made the `TypedArray` types [generic](https://devblogs.microsoft.com/typescript/announcing-typescript-5-7/#support-for---target-es2024-and---lib-es2024). The generic parameter has a default argument that should commonly spare callers from having to specify it. It does appear to be needed in this case, however. When I add it on the first array element in your playground, the code appears to transpile correctly.

```
  for (const TypedArray of [
    Uint8Array<ArrayBuffer>, <==
    Uint16Array,
    Uint32Array,
    ....
```

jhugman:
That's super interesting. Thanks for digging that out!

@Johennes As the reviewer and official typescript wonk :), would you prefer I add the type parameter (now breaks in v5.6.3),  or my dumb re-write that works in both?

zzorba (approved):
LGTM

jhugman:
Sorry @Johennes , I was over hasty merging this without responding to your comment.

> Do we deliberately support different TS versions? Could we not lock in a specific version by creating a lock file from package.json and then use that when running the tests?

We currently don't have *any* nodejs dependencies, just `devDependencies`, so yes, I think we could do this.

My understanding was that it wasn't a great idea [to keep a lock file in version control for libraries](https://github.com/yarnpkg/yarn/issues/838#issuecomment-253362537). I am very easily persuaded one way or the other, since a) the `yarn.lock` files for libraries seem to be ignored when building the app, b) this is only for tests (at the moment).

What about `Cargo.lock`? Well, `ubrn` is the application and we don't provide any Rust that runs in the user's apps, so I think we should have a `Cargo.lock`.

Johennes:
No worries at all. Merging this was the right thing to do as it unbreaks the CI. Sorry, I should have approved first and discussed potential future improvements second.

It's interesting that the person in that GitHub issue recommends not to include `yarn.lock` in version control when the official [yarn docs](https://classic.yarnpkg.com/lang/en/docs/yarn-lock/#toc-check-into-source-control) state the opposite. 😵‍💫

Now that we have no failures anymore, it's not a real problem. I've put in https://github.com/jhugman/uniffi-bindgen-react-native/issues/170 to discuss and possibly take action some time in the future.

PR #169, opened 2024-12-02
Add Github Actions to publish to Cocoapods and npm

### LINKED ISSUE #166: Publish to CocoaPods
We should: 
1. publish to CocoaPods, either as `UniffiBindgenReactNative` or `uniffi-bindgen-react-native`.
2. (one time) add `uniffi-bindgen-react-native` as a spec dependency in the template file.
3. publish to npmjs.
4. … Do this as a CI job.

So far, this issue is to publish to CocoaPods.

jhugman:
Fixes #166.

This is preparatory work to publish to the averred package managers.

Still to happen for first release:

- Add an `NPM_TOKEN` to npm.
- Tag the release as `0.28.3-0`
- Perform the release.

After the first release, we still need to:

- verify releases happen
- update the documentation about pre-releases.

Subsequent releases, we should think about automating:

- update the `package.json` and ``Cargo.toml` version numbers.
- rotate the `CHANGELOG.md` (currently we don't have a change log).
- tag the relase
- do the release


uniffi-bindgen-react-native.podspec
```diff
@@ -9,7 +9,11 @@ Pod::Spec.new do |s|
...
   s.author           = { package['author']['name'] => package['author']['email'] }
   s.source           = { :git => package['repository']['url'], :tag => s.version.to_s }
-  s.platform     = :ios, '9.0'
-  s.source_files  = 'cpp/includes/*.{h,cpp,hpp}'
-  s.dependency 'React'
+  s.platform         = :ios, '13.0'
+  s.source_files     = 'cpp/includes/*.{h,cpp,hpp}'
+  s.swift_versions   = '4.0'
+  s.pod_target_xcconfig = {
+    'SWIFT_OPTIMIZATION_LEVEL' => '-Onone',
```
Johennes, line 16: This struck me as slightly unexpected. Why do we disable compiler optimization here?

Johennes (approved):
This looks good as far as I can judge. Hard to test without a release of course. 🙈

uniffi-bindgen-react-native.podspec
jhugman: We don't use any Swift, and it was recommended by `pod lib lint`. I think if/when we start using Swift we should revisit this optimization.

PR #173, opened 2024-12-03
Add yarn.lock file to git

### LINKED ISSUE #170: Add yarn.lock to version control?
In https://github.com/jhugman/uniffi-bindgen-react-native/pull/168#issuecomment-2512391838 we discussed whether or not to add yarn's lock file to version control. Apparently there is different guidance on this online. https://github.com/yarnpkg/yarn/issues/838#issuecomment-253362537 advises against it whereas the [yarn (classic) docs](https://classic.yarnpkg.com/lang/en/docs/yarn-lock/#toc-check-into-source-control) say the opposite. Adding the lock file would enable reproducible tests but might have downsides that we should figure out.

jhugman:
Fixes #170 .


xtask/src/bootstrap/yarn.rs
```diff
@@ -37,7 +37,6 @@ impl Bootstrap for YarnCmd {
         let mut cmd = Command::new("yarn");
         run_cmd(
             cmd.current_dir(repository_root()?)
-                .arg("--no-lockfile")
```
Johennes, line 40: It might be good to use `--frozen-lockfile` which is usually recommended for CI, I believe, because in some cases `yarn install` can cause updates even if a lock file exists. Not a blocker though.

Johennes (approved):
LGTM

Thanks for tackling this so swiftly. 🚀

PR #174, opened 2024-12-04
Initial refactor in preparing for WASM

jhugman:
This PR does a small number of things, which makes it O(_n * m_).

Principally, it is to make room for the WASM bindings generator, with a `--flavor` ABI Flavor to switch between them.

This necessitated quite a large refactor. Most of this is mechanical; a lot of changes are files moving and `use` statements being updated.

The upshot of this PR is that we now have a `--flavor wasm` flag in the `xtask run`, which runs the typescript tests using `tsx`/`nodejs`.

The ReactNative bindings generator calls into a newly refactored `get_typescript` module which accepts a flavor.

Anticipating that this is not the last time we'll add a switch like flavor, this is contained in a `SwitchesArg` struct.


zzorba (approved):
LGTM, ran it locally without issue.

PR #175, opened 2024-12-07
Update "bin" of package.json to point to an executable

kraenhansen:
The file referenced by the `bin` property of a package.json, must be executable, meaning, it must start with the "node shebang". Or as the docs put it: https://docs.npmjs.com/cli/v10/configuring-npm/package-json#bin

> Please make sure that your file(s) referenced in bin starts with #!/usr/bin/env node, otherwise the scripts are started without the node executable!

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,8 +14,8 @@
   "type": "module",
   "main": "./typescript/src/index.ts",
   "bin": {
-    "ubrn": "./bin/cli.cjs",
-    "uniffi-bindgen-react-native": "./bin/cli.cjs"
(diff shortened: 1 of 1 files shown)


kraenhansen:
> Just to close the loop, change is working great for me locally.

_Originally posted by @zzorba in https://github.com/jhugman/uniffi-bindgen-react-native/issues/108#issuecomment-2526363102_

zzorba (approved):
LGTM, I don't remember why we added this in the first place, but happy to remove it now.

Issue #178, opened 2024-12-07
Definition of `uniffiIsDebug` yields "Cannot find name 'process'"


kraenhansen:
In a package locking down the `types` of its TypeScript configuration the `process` global won't be available and the code generated for `uniffiIsDebug` errors:

```
> tsc

src/generated/foobar.ts:72:23 - error TS2591: Cannot find name 'process'. Do you need to install type definitions for node? Try `npm i --save-dev @types/node` and then add 'node' to the types field in your tsconfig.

72 const uniffiIsDebug = process?.env?.NODE_ENV !== "production" || false;
                         ~~~~~~~


Found 1 error in src/generated/foobar.ts:72
```

![Screenshot 2024-12-07 at 16 49 02](https://github.com/user-attachments/assets/f70ccb04-dd47-4cbe-8cac-c00e26c0690f)

In the code I'm generating, this constant is not even used.

This is the tsconfig I'm using:

```json
{
  "extends": "@tsconfig/react-native",
  "compilerOptions": {
    "types": ["react-native"],
    "outDir": "dist",
    "noEmit": false,
    "declaration": true,
    "declarationMap": true,
    "sourceMap": true
  },
  "include": [
    "src"
  ]
}
```

jhugman:
Oh interesting.

PR #179, opened 2024-12-07
Make `uniffiIsDebug` definition runtime agnostic

kraenhansen:
A proposed fix for https://github.com/jhugman/uniffi-bindgen-react-native/issues/178 is to use `typeof` as this handles undefined variables gracefully and use a `@ts-ignore` comment to make the TS compiler happy 🙂

This fixes https://github.com/jhugman/uniffi-bindgen-react-native/issues/178.

### DIFF
--- a/crates/ubrn_bindgen/src/bindings/react_native/gen_typescript/templates/wrapper.ts
+++ b/crates/ubrn_bindgen/src/bindings/react_native/gen_typescript/templates/wrapper.ts
@@ -40,8 +40,12 @@ const {
 } = {{ entry.0.1 }}.converters;
 {%- endfor %}
 
-// @ts-ignore -- The process global might not be defined
-const uniffiIsDebug = (typeof process !== "object" || typeof process.env !== "object" || process.env.NODE_ENV !== "production" || {{ config.is_debug() }});
+const uniffiIsDebug =
+  // @ts-ignore -- The process global might not be defined
+  typeof process !== "object" ||
+  // @ts-ignore -- The process global might not be defined
+  process?.env?.NODE_ENV !== "production" ||
+  {{ config.is_debug() }};
 
 {%- call ts::docstring_value(ci.namespace_docstring(), 0) %}
 
--- a/crates/ubrn_bindgen/src/bindings/react_native/gen_typescript/templates/wrapper.ts
+++ b/crates/ubrn_bindgen/src/bindings/react_native/gen_typescript/templates/wrapper.ts
@@ -40,7 +40,8 @@ const {
 } = {{ entry.0.1 }}.converters;
 {%- endfor %}
 
-const uniffiIsDebug = (process?.env?.NODE_ENV !== "production" || {{ config.is_debug() }});
+// @ts-ignore -- The process global might not be defined
+const uniffiIsDebug = (typeof process !== "object" || typeof process.env !== "object" || process.env.NODE_ENV !== "production" || {{ config.is_debug() }});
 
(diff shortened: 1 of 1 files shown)


crates/ubrn_bindgen/src/bindings/react_native/gen_typescript/templates/wrapper.ts
```diff
@@ -40,7 +40,8 @@ const {
 } = {{ entry.0.1 }}.converters;
 {%- endfor %}
 
-const uniffiIsDebug = (process?.env?.NODE_ENV !== "production" || {{ config.is_debug() }});
+// @ts-ignore -- The process global might not be defined
+const uniffiIsDebug = (typeof process !== "object" || typeof process.env !== "object" || process.env.NODE_ENV !== "production" || {{ config.is_debug() }});
```
jhugman, line 44: Does this mean if _either_ `process` or `process.env` are not an object, OR or `process.env.NODE_ENV` is not `"production"`, then `uniffiIsDebug` is set to `true`?

I'm not sure I understand the difference between this and the nullish coelecsing (`?.`) it replaces. If there is none,  I _think_ the intent of the `?.` version is more clear. What do you think?

kraenhansen: I've tried to match the semantics of the existing code.
I interpret this as: If both the `process` and `process.env` are objects the `NODE_ENV` must not be `"production"` for `uniffiIsDebug` to remain true.

> the difference between this and the nullish coelecsing (?.) it replaces

Using the `typeof` operator, you can provide an undefined identifier whereas the nullish coalescing operator expects the first variable to be (at least) defined in the scope - here's an example from a node REPL:

```
> missing?.env?.NODE_ENV === "production"
Uncaught ReferenceError: missing is not defined
> typeof missing === "object"
false
```

jhugman: Ah ok, I understand now. Thank you for being patient with me.

jhugman (approved):

crates/ubrn_bindgen/src/bindings/react_native/gen_typescript/templates/wrapper.ts
kraenhansen: Hmm - I just ran this locally and prettier is messing up the placement of the `// @ts-ignore` comment.
I'll push another commit to fix that.

Johennes:
The failures in the compat jobs (here and on the other PRs) are due to a parameter change in upstream create-react-native-library. Hoping that https://github.com/jhugman/uniffi-bindgen-react-native/pull/182 should fix it but I think this doesn't have to be a blocker.

crates/ubrn_bindgen/src/bindings/react_native/gen_typescript/templates/wrapper.ts
```diff
@@ -40,7 +40,12 @@ const {
 } = {{ entry.0.1 }}.converters;
 {%- endfor %}
 
-const uniffiIsDebug = (process?.env?.NODE_ENV !== "production" || {{ config.is_debug() }});
+const uniffiIsDebug =
+  // @ts-ignore -- The process global might not be defined
+  typeof process !== "object" ||
+  // @ts-ignore -- The process global might not be defined
+  process?.env?.NODE_ENV !== "production" ||
+  {{ config.is_debug() }};
```
jhugman, line 48: :chefs-kiss: Perfect thank you.

PR #182, opened 2024-12-09
Handle type parameter change in crnl 0.45.1

Johennes:
Looks like this was renamed in https://github.com/callstack/react-native-builder-bob/commit/e9f80ea0055f5b3b7d123dcd08b3de8052206c92 which now breaks some of the compat jobs.

https://github.com/jhugman/uniffi-bindgen-react-native/actions/runs/12208165279/job/34060907217

### DIFF
--- a/scripts/test-turbo-modules.sh
+++ b/scripts/test-turbo-modules.sh
@@ -174,6 +174,11 @@ create_library() {
     rm -rf "$base" || error "Failed to remove existing directory $base"
   fi
 
+  local type="turbo-module"
+  if [[ "$BOB_VERSION" != "0.45.1" && "$BOB_VERSION" = "`echo -e "0.45.1\n$BOB_VERSION" | sort -V | head -n1`" ]]; then
+    type="module-new"
+  fi
+
   echo "-- Creating library $PROJECT_SLUG with create-react-native-library@$BOB_VERSION"
   npm_config_yes=true npx "create-react-native-library@$BOB_VERSION" \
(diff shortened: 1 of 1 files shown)


scripts/test-turbo-modules.sh
```diff
@@ -174,6 +174,11 @@ create_library() {
     rm -rf "$base" || error "Failed to remove existing directory $base"
   fi
 
+  local type="turbo-module"
+  if [[ "$BOB_VERSION" != "0.45.1" && "$BOB_VERSION" = "`echo -e "0.45.1\n$BOB_VERSION" | sort -V | head -n1`" ]]; then
```
kraenhansen, line 178: This is probably fine for now, but this breaks once builder-bob v0.100.0 ships 😊

If this was TypeScript we could have used the `server` package to do this reliably, if you want to make this more robust you could also add it as a dev-dependency and call into `node` here.

jhugman: I'm going to delegate my review approval power to @kraenhansen who seems to be better placed to approve or disapprove.

I would suggest, however, _not_ translating this script into TS/JS, but do it in a different PR.

jhugman (approved):

scripts/test-turbo-modules.sh
kraenhansen: You're right - I didn't realize [the meaning of that](https://www.gnu.org/software/coreutils/manual/html_node/Version-sort-overview.html) 👍

PR #183, opened 2024-12-09
Fix CLI working without an extension

zzorba:


kraenhansen:
It would be sad if the package can't be ESM, just because of this 🤔

jhugman (approved):
This looks fine. 

You should also add the `.cjs` to https://github.com/jhugman/uniffi-bindgen-react-native/blob/main/scripts/test-turbo-modules.sh#L71 and https://github.com/jhugman/uniffi-bindgen-react-native/blob/main/bin/cli.cjs#L35.

I'll approve this now so as not to wait for another round of feedback.

kraenhansen:
LGTM! :shipit:

PR #184, opened 2024-12-09
Use version released to Cocoapods and npm

jhugman:
Follow up to #166, now that has been released.

Once this is landed, we should:

1. land anything that is to be in the next release.
2. do a release.

We need to do a release now because:
* The generated podspec file now includes a dependency on `uniffi-bindgen-react-native`.
* We can remove pre-release caveats from the documentation.

### DIFF
--- a/crates/ubrn_cli/src/codegen/templates/module-template.podspec
+++ b/crates/ubrn_cli/src/codegen/templates/module-template.podspec
@@ -32,6 +32,9 @@ Pod::Spec.new do |s|
   {%- let bindings = self.relative_to(root, dir) -%}
   s.source_files = "{{ ios }}/**/*.{h,m,mm}", "{{ codegen }}/**/*.{h,m,mm}", "{{ tm }}/**/*.{hpp,cpp,c,h}", "{{ bindings }}/**/*.{hpp,cpp,c,h}"
   s.vendored_frameworks = "{{ framework }}"
+  if ENV['UBRN_INSTALLED_LOCALLY'] != '1' then
+    s.dependency    "uniffi-bindgen-react-native", "{{ self.config.project.ubrn_version() }}"
+  end
 
   # Use install_modules_dependencies helper to install the dependencies if React Native version >=0.71.0.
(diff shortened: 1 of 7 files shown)


crates/ubrn_cli/src/codegen/templates/module-template.podspec
```diff
@@ -32,6 +32,7 @@ Pod::Spec.new do |s|
   {%- let bindings = self.relative_to(root, dir) -%}
   s.source_files = "{{ ios }}/**/*.{h,m,mm}", "{{ codegen }}/**/*.{h,m,mm}", "{{ tm }}/**/*.{hpp,cpp,c,h}", "{{ bindings }}/**/*.{hpp,cpp,c,h}"
   s.vendored_frameworks = "{{ framework }}"
+  s.dependency    "uniffi-bindgen-react-native", "{{ self.config.project.ubrn_version() }}"
```
Johennes, line 35: I was briefly considering whether we should do something to support a local development flow where the pod is loaded from node_modules. However, I guess it's a rare use case and most likely not worth the hassle.

Johennes (approved):
LGTM

zzorba (approved):
LGTM

crates/ubrn_cli/src/codegen/templates/module-template.podspec
jhugman: If local development means: local development of ubrn, I spent a bunch of time yesterday trying to get the local development flow going. 

Where I ended up was: getting the tests to comment out this line. 

If local development means: local development of the subject rn-rust library, I wasn't really concerned about that, given that it's only adding the `cpp/include/*.h` files from ubrn, which—to a first approximation—don't really change. That and caching in a the `Pods` directory, means that it's as local as any other dependency.

Johennes: I kind of meant a mix of these two. I'm sometimes testing ubrn from a SHA against react-native-matrix-sdk. But it's exotic enough a use case that I can just make the small changes locally myself. So I think this is fine. 👍

PR #185, opened 2024-12-09
Make first run more informative while compiling

### LINKED ISSUE #181: Less silence from the CLI when ran the first time
Since the CLI is calling into [cargo run](https://github.com/jhugman/uniffi-bindgen-react-native/blob/56a74a521f9294af77ace1bcf44eb707ccf38b5e/bin/cli.cjs#L43) it takes quite a bit of time on the initial call (as dependencies needs to be downloaded and CLI has to be built). While I agree that I don't want this feedback on every call and passing the `--quite` flag makes sense in that case, it would be great to get some sort of feedback when called initially.

A naive solution would be to check for the existence of the `target` directory and log something about "Building the CLI".

jhugman:
Fixes #181 

This uses the "naive" solution suggested by @kraenhansen—it just detects the absence of the `target` directory.

I added the `--quiet` flag sometime ago because the default cargo output of the `ubrn`  was getting to be annoying. Hopefully this is a reasonable compromise.

### DIFF
--- a/bin/cli.cjs
+++ b/bin/cli.cjs
@@ -39,6 +39,12 @@ if (args.length === 1 && args[0] === "--path") {
 // Construct the path to the Cargo.toml file
 const manifestPath = path.join(rootDir, "crates/ubrn_cli/Cargo.toml");
 
+if (!fs.existsSync(path.join(rootDir, "target"))) {
+  console.log(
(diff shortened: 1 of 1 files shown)


Johennes:
:shipit:

Johennes (approved):

PR #187, opened 2024-12-16
Switch from ArrayBuffer to ByteArray

jhugman:

ubrn lowers most types to an `ArrayBuffer` and then maps it to a `jsi::ArrayBuffer`, and then on to `RustBuffer`.
`wasm-bindgen` maps a `Uint8Array` on to a `Vec<u8>`.

In order to lower types to a `Uint8Array` for WASM and `ArrayBuffer` for JSI, some amount of bundler violence was needed. Given we're using multiple bundlers to package up the frontend Typescript, _and_ we're not really in control of what bundler the developer uses, this doesn't seem like a viable approach.

Instead, this PR changes ubrn to lower types down to a `Uint8Array`, so that both WASM and React Native use the same underlying representation of byte arrays.

Unfortunately, jsi doesn't expose API for typed arrays (it may in the future), so we look up the [`buffer` property](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray#instance_properties) to get an underlying `ArrayBuffer`.

On the way back, an object of type `{ buffer: ArrayBuffer; }` is returned, which is enough to make the uniffi work and the typescript compiler think it that a typed array has come back from Rust/C++.

Most of this PR was mechanical refactoring (e.g. `ArrayBuffer` renamed to `UniffiByteArray`) but there are enough manual changes which merit it being O(_n_).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,30 @@
+# Upcoming releases
+
+## 🦊 What's Changed
+* Switched from passing `ArrayBuffer`s to using `Uint8Array`, to accommodate WASM better. ([#187](https://github.com/jhugman/uniffi-bindgen-react-native/pull/187))
+
+**Full Changelog**: https://github.com/jhugman/uniffi-bindgen-react-native/compare/0.28.3-1...main
+
+# 0.28.3-1
+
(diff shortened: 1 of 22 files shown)


zzorba (approved):
LGTM

This is probably something we should document in the changelog on the next release, given there's a type change?  Should we make a spot to accumulate those?

PR #188, opened 2024-12-16
Arithmetic working with WASM!

jhugman:
This is the first cut of a Rust generator, enough to run the `fixture/arithmetic`.

Unfortunately due to https://github.com/mozilla/uniffi-rs/pull/2329 not yet released, we have to introduce a procmacro version of the fixture.

In this PR:

1. Introduce a Rust crate generator; this uses 
2. a runtime which uses typealiasing and import renaming to use `wasm-bindgen` indirectly in the generated code.
3. the generated code needs a CallStatus to be defined in Rust; this differs from the JSI version where the CallStatus is defined in typescript. So the `rust-call.ts` module is made into a class with three methods instead of three functions.
4. the generated Rust is then compiled and put through wasm-bindgen-cli where it generates a low-level JS file.
5. the high level typescript now uses the flavor to decide whether it uses the JSI generated typescript or the WASM, and which implementation of the CallStatus.
6. Uses `tsx` to run the whole thing.


zzorba (approved):
LGTM, huge progress!

PR #189, opened 2024-12-17
Remove reference to name field in the ubrn.config.yaml docs

jhugman:

This is a documentation only change, based on troubleshooting with Annie.


docs/src/reference/config-yaml.md
```diff
@@ -162,3 +162,13 @@ In this case, changing the location of the `ts` directory will require changing
...
+
+For example, if you want to add C++ files to the library, you may want to change the build files.
+
+```yaml
+noOverwrite:
+    - "*.podspec"
+    - CMakeLists.txt
+```
+
+You can generate the build files once then not overwrite them. Once exclude the files they can be safely edited.
```
Johennes, line 174:
```suggestion
You can generate the build files once then not overwrite them. Once you excluded the files, they can be safely edited.
```

Johennes (approved):
Looks good to me. 👍

PR #191, opened 2024-12-18
Coverall2 and Rondpoint working with WASM!

jhugman:
This PR adds:

- lowlevel structures
  - RustBuffers, 
  - Floats and 
  - c_void Void Pointers.
- high level mechanisms to do in a browser context:
  - encoding and decoding strings, using `TextEncoder` and `TextDecoder`.
  - tie Objects into the Garbage Collector with a `FinalizationRegistry`.

This is in service of getting the `rondpoint` fixture running.

Several things to note here:

### `error: rondpointpm::Optionneur is a private type`

Structs which derive the `uniffi::Object` currently need to have visibility `pub`. If they're not, then errors like: `error: rondpointpm::Optionneur is a private type` are shown. 

This is due to `uniffi-rs` generating a public C ABI which uses the private struct as part of its trait bounds. This feels like it's a Rust bug (I hate saying that). 

Next steps:

- File an issue: 
  - `uniffi-rs` could generate a simpler ABI which didn't contain the private type.
  - `uniffi-rs` could prevent non-`pub` structs
- Document: `uniffi-bindgen-react-native` should add this to a trouble-shooting page.

### Additing a procmacro version of rondpoint and coverall2

This is due to https://github.com/mozilla/uniffi-rs/pull/2329 not yet being release.

Since we have now released to CocoaPods and npm, I can build against uniffi-main. This is o k a y for a little project like this, but I do worry about that a chem-spill requiring a rapid fix and release may get messy.

### Default attribute values not seeming to work in proc-macro `uniffi::Record`

I don't seem to be able to get default attributes working for records going: as in the typescript isn't getting generated. If this is a bug, this is almost certainly a uniffi-rs bug. Will file.

### DIFF
--- a/crates/ubrn_bindgen/src/bindings/extensions.rs
+++ b/crates/ubrn_bindgen/src/bindings/extensions.rs
@@ -300,6 +300,11 @@ pub(crate) impl FfiFunction {
         let name = self.name();
         name.contains("ffi__") && name.contains("_internal_")
     }
+
+    fn is_unsafe(&self) -> bool {
+        let name = self.name();
+        name.contains("_fn_clone_") || name.contains("_fn_free_") || name.contains("_rustbuffer_")
+    }
 }
 
 #[ext]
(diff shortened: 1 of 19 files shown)


zzorba (approved):
LGTM, huge progress

Issue #190, opened 2024-12-18
Allow specifying the profile for cargo build


Johennes:
Following through from https://github.com/matrix-org/matrix-rust-sdk/issues/4416, matrix-rust-sdk uses the `reldbg` profile by [default](https://github.com/matrix-org/matrix-rust-sdk/blob/main/xtask/src/swift.rs#L70). This cannot be selected via UBRN though because we only have a `--release` flag that [switches](https://github.com/jhugman/uniffi-bindgen-react-native/blob/main/crates/ubrn_common/src/rust_crate.rs#L25) between the `debug` and `release` profiles.

Adding `--profile reldbg` via `cargoExtras` does not work because UBRN will still look for the library under the `debug` or `release` folder.

Proposal: Add a `profile` config / CLI option to select the specific profile. If both `profile` and `release` are specified, let `profile` take precedence ([as in matrix-rust-sdk](https://github.com/matrix-org/matrix-rust-sdk/blob/main/xtask/src/swift.rs#L31)).

jhugman:
This is a good idea, thanks for suggesting it!

It looks like it shouldn't be too difficult to add.

The [`CommonBuildArgs` struct](https://github.com/jhugman/uniffi-bindgen-react-native/blob/0ea522e614621b56cb9ae4af508deccc74b30d52/crates/ubrn_cli/src/building.rs#L84-L107).

Adding an additional `profile` attribute:

```rs
    /// Suitable documentation
    #[clap(long, conflicts_with_all = ["release"]]
    profile: Option<String>,
```

and adapting the existing `fn profile(&self)` might be all you need. 

I checked: the [android `cargo ndk` command](https://github.com/jhugman/uniffi-bindgen-react-native/blob/0ea522e614621b56cb9ae4af508deccc74b30d52/crates/ubrn_cli/src/android.rs#L235-L241) needs to change, as does [the `cargo build` for iOS](https://github.com/jhugman/uniffi-bindgen-react-native/blob/0ea522e614621b56cb9ae4af508deccc74b30d52/crates/ubrn_cli/src/ios.rs#L210-L218).

I'm unlikely to get to this before next month, however. Would you like to pick this up, or would you like to wait?

Johennes:
Thanks for the pointers. ❤️ 

I took a stab at it in https://github.com/jhugman/uniffi-bindgen-react-native/pull/192.

PR #192, opened 2024-12-19
Add --profile build argument

Johennes:
Fixes: #190

### DIFF
--- a/crates/ubrn_cli/src/android.rs
+++ b/crates/ubrn_cli/src/android.rs
@@ -232,7 +232,8 @@ impl AndroidArgs {
             .arg(target.to_string())
             .arg("--platform")
             .arg(format!("{}", api_level));
-        if self.common_args.profile() != "release" {
+        let profile = self.common_args.profile();
+        if profile != "release" {
             cmd.arg("--no-strip");
         }
         cmd.arg("--").arg("build");
--- a/crates/ubrn_cli/src/android.rs
+++ b/crates/ubrn_cli/src/android.rs
@@ -235,10 +235,10 @@ impl AndroidArgs {
         if self.common_args.profile() != "release" {
             cmd.arg("--no-strip");
         }
-        cmd.arg("--")
-            .arg("build")
-            .arg("--profile")
-            .arg(self.common_args.profile());
+        cmd.arg("--").arg("build");
+        if self.common_args.profile() != "debug" {
+            cmd.arg("--profile").arg(self.common_args.profile());
+        }
         cmd.args(cargo_extras.clone());
         run_cmd(cmd.current_dir(rust_dir))?;
         Ok(target.clone())
--- a/crates/ubrn_cli/src/android.rs
+++ b/crates/ubrn_cli/src/android.rs
@@ -236,8 +236,8 @@ impl AndroidArgs {
             cmd.arg("--no-strip");
         }
         cmd.arg("--").arg("build");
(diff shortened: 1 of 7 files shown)


crates/ubrn_cli/src/ios.rs
```diff
@@ -213,8 +213,8 @@ impl IOsArgs {
             .arg(manifest_path)
             .arg("--target")
             .arg(&target.triple);
-        if self.common_args.release {
-            cmd.arg("--release");
+        if self.common_args.profile() != "debug" {
+            cmd.arg("--profile").arg(self.common_args.profile());
```
jhugman, line 217: Nit: extract a local variable
```suggestion
        let profile = self.common_args.profile();
        if profile != "debug" {
            cmd.args(["--profile", profile]);
```

crates/ubrn_cli/src/android.rs
```diff
@@ -232,12 +232,12 @@ impl AndroidArgs {
...
             .arg(format!("{}", api_level));
-        if !self.common_args.release {
+        if self.common_args.profile() != "release" {
             cmd.arg("--no-strip");
         }
         cmd.arg("--").arg("build");
-        if self.common_args.release {
-            cmd.arg("--release");
+        if self.common_args.profile() != "debug" {
+            cmd.arg("--profile").arg(self.common_args.profile());
```
jhugman, line 240: Nit: use local variable
```suggestion
        if profile != "debug" {
            cmd.args(["--profile", profile]);
```

crates/ubrn_cli/src/android.rs
```diff
@@ -232,12 +232,12 @@ impl AndroidArgs {
             .arg(target.to_string())
             .arg("--platform")
             .arg(format!("{}", api_level));
-        if !self.common_args.release {
+        if self.common_args.profile() != "release" {
```
jhugman, line 235: Nit: extract a local variable
```suggestion
        let profile = self.common_args.profile();
        if profile != "release" {
```

xtask/src/run/rust_crate.rs
```diff
@@ -21,6 +21,12 @@ pub(crate) struct CrateArg {
     #[clap(long, requires = "crate_dir", default_value = "false")]
     pub(crate) release: bool,
 
+    /// Use a specific build profile
+    ///
+    /// This overrides the release flag if both are specified.
+    #[clap(long, requires = "crate_dir")]
```
jhugman, line 27: This can be a bit fiddly sometimes, but I think adding `conflicts_with_all` to either this attribute or `release` should make `--profile` and `--release` mutually exclusive.

```suggestion
    #[clap(long, requires = "crate_dir", conflicts_with_all = ["release"])]
```

jhugman (approved):
@Johennes You're just so quick! Thank you so much.

I've approved this with nits. I don't think they're big enough to wait for another review cycle, but I would like these fixed before landing.

Also: would you mind adding an entry to the `CHANGELOG.md`? I think this qualifies as "✨ What's New"

xtask/src/run/rust_crate.rs
Johennes: Oh, nice! I wasn't aware of this feature. 🤯

Johennes:
> Also: would you mind adding an entry to the CHANGELOG.md? I think this qualifies as "✨ What's New"

Oh, I completely forgot that we have a changelog now. 🤦‍♂️

PR #197, opened 2024-12-22
Prevent TypeScript from checking generated files for errors

hassankhan:
This PR adds `@ts-nocheck` to the top of the generated TypeScript output files, which stops tsc from typechecking them.

https://www.typescriptlang.org/docs/handbook/intro-to-js-ts.html#ts-check


Johennes (approved):
Thanks for submitting this. I feel like it might not hurt to run _some_ checks on the generated code. However, until we've figured out which ones those might be, this feels like a good step as it prevents unexpected headaches for consumers.

PR #200, opened 2024-12-27
Stabilise `require.resolve` by looking up `package.json` instead of entrypoint

hassankhan:
This change decouples the path lookup from the internal entrypoint of `uniffi-bindgen-react-native`.

Since `package.json` always lives at the root of an npm module, `require.resolve`-ing it and then either stripping `/package.json` from the end, or passing the path to `path.dirname()` will get the module's path on-disk.

### DIFF
--- a/bin/cli.cjs
+++ b/bin/cli.cjs
@@ -24,8 +24,10 @@ function getRootDir() {
   }
 
   // Fallback to using require.resolve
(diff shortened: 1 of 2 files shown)


jhugman (approved):
This is really good. Thank you for splitting it out to a new PR, it makes it so much easier to review and land.

PR #202, opened 2025-01-06
Use the C-ABI to call into Rust crates for WASM!

jhugman:
This PR started off as getting errors working.

This, turned out, fairly straightforward. However, showing that the fixed worked by enabling a test fixture turned out to be. less straightforward.

This led to a re-write of the rust generation. In this PR, we switch from using a Rust-to-Rust calls which happen to be exposed by uniffi as `extern "C"` to relying on the ABI and only the ABI.

This means we do not have to rely on Rust visibility rules, meaning that a large number of fixtures now Just Work™, without needing to wait for uniffi-rs to release `v0.29.x`.

We _don't_ need to keep track of the module paths, so I am hoping that multi-module crates should be straightforward. The only reason why the `ext-types` fixture is not enabled is because the `uniffi_one.udl` has a callback interface.

I don't think this approach is compatible with a some of the optimizations outlined in https://github.com/mozilla/uniffi-rs/issues/2331, but I think this is a good approach for 0.28.


crates/ubrn_common/src/rust_crate.rs
```diff
@@ -88,6 +88,31 @@ impl CrateMetadata {
     }
```
jhugman, line 88: This refactor used to be relevant to this PR, but now we're using `ci.crate_name()` it is not. 

It is still a useful refactor, but we're in a high velocity phase of development, so would rather not spend time picking it out into a separate PR.

WDYT? should I leave it here, or break it out?

jhugman (commented):
Self review.

crates/ubrn_common/src/rust_crate.rs
zzorba: My opinion, leave it in.  

I've generally found that trying to separate out changes after the fact leads to more problems than not, but ultimately your call.

zzorba (approved):
LGTM

zzorba:
Sorry, realized this was a dangling PR I had finished but not hit submit on.

PR #205, opened 2025-01-20
Callbacks now have UniffiResult to communicate between typescript and C++

jhugman:
This is the first of between 1 and 3 PRs on the JSI to unify `uniffi_out_return` and `RustCallStatus` into new `UniffiResult<T>` object.

This PR eliminates the need for out parameters for callbacks between typescript and C++. The out paramaters are still needed to call into Rust, but these are all now hidden from the typescript.

Primarily this is to enable WASM callbacks.

This is not yet done for ordinary functions (which use an out parameter `RustCallStatus` for errors), and may never be.

For async callbacks, the errors and returns are passed by calling from typescript into Rust. There is a complication about __what__ is being called, but I don't think this needs any action yet.

For Typescript to Rust promises/futures, there is already an result object of sorts: it is monomorphisized already, which we're going to have to do soon with `UniffiResult`, so I'm not quite sure if we need to use those instead of a new type.

### DIFF
--- a/crates/ubrn_bindgen/src/bindings/extensions.rs
+++ b/crates/ubrn_bindgen/src/bindings/extensions.rs
@@ -384,6 +384,13 @@ pub(crate) impl FfiCallbackFunction {
         !self.is_future_callback() || self.name() == "RustFutureContinuationCallback"
     }
 
+    fn returns_result(&self) -> bool {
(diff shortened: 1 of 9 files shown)


zzorba (approved):
LGTM

PR #207, opened 2025-01-21
Adjust template to allow for hot reload via metro of running apps

zzorba:
Won't apply if a rust change is made, but a rebuild is required then anyway.


jhugman (approved):
LGTM. It would be nice to have a comment for why these flags are in the template, but I wouldn't hold up for another review cycle.

PR #208, opened 2025-01-23
Fixup GC'ing objects with callbacks intermittent crasher

jhugman:

Fixes https://github.com/zzorba/ubrn-sample-app/pull/1

The investigation for this was in https://github.com/zzorba/ubrn-sample-app/pull/1.

The key insight here is that the "fake GC" destructor is calling into Rust which then calls a free callback in hermes, which somehow triggers another GC.

The fix is to schedule the free callback on the event loop.

This also fixes previously encountered `"Stats not submitted."` assertion failures.

Possible other things to do for this: document that you shouldn't call callbacks in a `Drop::drop` though I'm not sure how to phrase that without being too restrictive.

### DIFF
--- a/crates/ubrn_bindgen/src/bindings/extensions.rs
+++ b/crates/ubrn_bindgen/src/bindings/extensions.rs
@@ -414,7 +414,7 @@ pub(crate) impl FfiCallbackFunction {
     }
 
     fn is_blocking(&self) -> bool {
-        self.name() != "RustFutureContinuationCallback"
+        !self.is_free_callback() && self.name() != "RustFutureContinuationCallback"
     }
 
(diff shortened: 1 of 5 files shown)


crates/ubrn_bindgen/src/bindings/extensions.rs
```diff
@@ -414,7 +414,7 @@ pub(crate) impl FfiCallbackFunction {
     }
 
     fn is_blocking(&self) -> bool {
-        self.name() != "RustFutureContinuationCallback"
+        !self.is_free_callback() && self.name() != "RustFutureContinuationCallback"
```
jhugman, line 417: As suspected, it's a one line fix.

PR #209, opened 2025-01-23
Followup to using non-blocking calls more widely

jhugman:
This is a followup to #208 to tidy up the `FfiCallbackFuncrtion` extensions in `extensions.rs`.

Most notably, it generalizes the `is_blocking` function to callback functions that do not have a return and are not expected to error, ever.


zzorba (approved):
LGTM

PR #211, opened 2025-01-27
Split compat job by platform and version

Johennes:
This is an attempt at splitting the "Compat" job so that we can report on it more granularly instead of having a single "Compat | failing" badge in the README.


jhugman (commented):
Oh yes, I very much like this. This will mean we can put more badges on the README.

Thank you!

Johennes:
> Oh yes, I very much like this. This will mean we can put more badges on the README.

Yes, exactly. It does require some duplication but it seemed better than having one version make the single badge red.

Also, splitting the historic versions (the ones we consciously choose to support) from the latest versions (which can break any time) seemed sensible.

Johennes:
The current changes will look like this in the README:

Screenshot 2025-01-27 at 15 36 14

jhugman (approved):
👨‍🍳 🤌 🚀 🚢

Issue #149, opened 2024-10-30
Generate Kotlin and Swift at the same time as generating TS/C++


jhugman:
@zzorba has been asking for this for a lilttle while.

While the main React Native app requires typescript and C++, iOS extensions still require Swift. 

Can we generate the Swift from the Rust at the same time as generating the Typescript and C++?

This issue just covers the generating and building of the Kotlin and Swift generation.

@almost adds this question: can we run the Kotlin and Swift bindings at the same time as the Typescript? This enables a host of new capabilities.

jhugman:
> I believe you can run the bindings at the same time, though I do not know if they share the process state or not (I would assume they would, but I'm just not sure since it hasn't been our need).

Interesting.

Eyeballing the code, I was under the impression that just generating the extra bindings was necessary but not sufficient for running in the same process. I am expecting that this would require changes to how [callback interfaces](https://github.com/mozilla/uniffi-rs/blob/main/uniffi_macros/src/export/callback_interface.rs) work.

Still, that is for another issue :)

Johennes:
Don't app extensions on iOS run in their own process (and with limited system resources) anyway? From what I know they can only share certain storage locations with the main app but not the same system process.

zzorba:
Yes that's correct about extensions, they are isolated and only share storage and keychain access.

But theoretically you could have swift code and react native code in the library both talking to the same binary.

Not sure if it's a large use case though?

Johennes:
Oh, I see what you mean now. Hm, yeah I suppose there could be a use case when you want to combine the bindings with some native APIs without having to poke out to TS and back in to another library.

Johennes:
@jhugman I'm curious about your thoughts on how to best integrate this into ubrn, starting with the Swift side. We could just call `generate_bindings` with `uniffi_bindgen::bindings::SwiftBindingGenerator` in [`cli.rs`](https://github.com/jhugman/uniffi-bindgen-react-native/blob/main/crates/ubrn_bindgen/src/cli.rs#L148). The Swift bindings generator produces headers and module maps that need to go into the `.xcframework`, however, and that has already been assembled at this point as part of the `build ios` command.

The build command currently has a [`no_xcodebuild`](https://github.com/jhugman/uniffi-bindgen-react-native/blob/main/crates/ubrn_cli/src/ios.rs#L126) to allow constructing the framework externally. I'm wondering if we could lift this and instead move the framework assembly into the `generate` command?

[...]

PR #214, opened 2025-01-31
Add option to generate native swift bindings

Johennes:
Relates to: #149

This adds an option to the `build ios` command to produce native Swift bindings.

Testing with matrix-rust-sdk, using this option would produce the `.h` and `.modulemap` files inside the `.xcframework` and the Swift files in the separate `swift` folder.

```shell
$ tree build/RnMatrixRustSdk.xcframework && tree swift
build/RnMatrixRustSdk.xcframework
├── Info.plist
├── ios-arm64
│   ├── Headers
│   │   ├── matrix_sdkFFI.h
│   │   ├── matrix_sdk_baseFFI.h
│   │   ├── matrix_sdk_commonFFI.h
│   │   ├── matrix_sdk_cryptoFFI.h
│   │   ├── matrix_sdk_ffiFFI.h
│   │   ├── matrix_sdk_uiFFI.h
│   │   └── module.modulemap
│   └── libmatrix_sdk_ffi.a
└── ios-arm64-simulator
    ├── Headers
    │   ├── matrix_sdkFFI.h
    │   ├── matrix_sdk_baseFFI.h
    │   ├── matrix_sdk_commonFFI.h
    │   ├── matrix_sdk_cryptoFFI.h
    │   ├── matrix_sdk_ffiFFI.h
    │   ├── matrix_sdk_uiFFI.h
    │   └── module.modulemap
    └── libmatrix_sdk_ffi.a

5 directories, 17 files
swift
├── matrix_sdk.swift
├── matrix_sdk_base.swift
├── matrix_sdk_common.swift
├── matrix_sdk_crypto.swift
├── matrix_sdk_ffi.swift
└── matrix_sdk_ui.swift

1 directory, 6 files
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Upcoming releases
 
 [//]: # (## ✨ What's New)
+
+* Add option to generate native swift bindings ([#214](https://github.com/jhugman/uniffi-bindgen-react-native/pull/214))
+
 [//]: # (## 🦊 What's Changed)
 [//]: # (## ⚠️ Breaking Changes)
 [//]: # (**Full Changelog**: https://github.com/jhugman/uniffi-bindgen-react-native/compare/{{previous}}...{{current}})
--- a/crates/ubrn_cli/src/codegen/mod.rs
+++ b/crates/ubrn_cli/src/codegen/mod.rs
@@ -479,6 +479,7 @@ mod tests {
                 targets: Default::default(),
                 cargo_extras: ExtraArgs::default(),
                 codegen_output_dir: "ios/generated".to_string(),
+                native_bindings_dir: "swift".to_string(),
             };
             let bindings = BindingsConfig {
                 cpp: "cpp/bindings".to_string(),
--- a/crates/ubrn_cli/src/codegen/mod.rs
+++ b/crates/ubrn_cli/src/codegen/mod.rs
@@ -479,7 +479,6 @@ mod tests {
                 targets: Default::default(),
                 cargo_extras: ExtraArgs::default(),
                 codegen_output_dir: "ios/generated".to_string(),
-                native_bindings_dir: "swift".to_string(),
             };
             let bindings = BindingsConfig {
                 cpp: "cpp/bindings".to_string(),
--- a/crates/ubrn_cli/src/codegen/templates/module-template.podspec
+++ b/crates/ubrn_cli/src/codegen/templates/module-template.podspec
@@ -30,7 +30,7 @@ Pod::Spec.new do |s|
   {%- let tm = self.relative_to(root, dir) %}
   {%- let dir = self.config.project.bindings.cpp_path(root) %}
   {%- let bindings = self.relative_to(root, dir) -%}
-  s.source_files = "{{ ios }}/**/*.{h,m,mm}", "{{ codegen }}/**/*.{h,m,mm}", "{{ tm }}/**/*.{hpp,cpp,c,h}", "{{ bindings }}/**/*.{hpp,cpp,c,h}"
+  s.source_files = "{{ ios }}/**/*.{h,m,mm,swift}", "{{ codegen }}/**/*.{h,m,mm}", "{{ tm }}/**/*.{hpp,cpp,c,h}", "{{ bindings }}/**/*.{hpp,cpp,c,h}"
   s.vendored_frameworks = "{{ framework }}"
   s.dependency    "uniffi-bindgen-react-native", "{{ self.config.project.ubrn_version() }}"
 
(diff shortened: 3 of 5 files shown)


Johennes:
This doesn't currently include any tests. We could possibly generate a `Package.swift` together with the bindings and try to build the Swift package from it. For my matrix-rust-sdk example, the following `Package.swift`

```
// swift-tools-version: 5.5

import PackageDescription

let package = Package(
    name: "MatrixRustSDK",
    platforms: [
        .iOS(.v15)
    ],
    products: [
        .library(name: "MatrixRustSDK", targets: ["MatrixRustSDK"]),
    ],
    targets: [
        .binaryTarget(name: "MatrixSDKFFI", path: "build/RnMatrixRustSdk.xcframework"),
        .target(name: "MatrixRustSDK", dependencies: [.target(name: "MatrixSDKFFI")], path: "swift")
    ]
)
```

can be built with

```
xcodebuild -scheme MatrixRustSDK -destination 'platform=iOS Simulator,OS=18.0,name=iPhone 16 Pro Max'
```

after generating the native bindings.

jhugman:
```
xcodebuild -scheme MatrixRustSDK -destination 'platform=iOS Simulator,OS=18.0,name=iPhone 16 Pro Max'
```

Doesn't this already get run as part of the iOS Compat tests?

I may be missing something but can we generate the `Package.swift` file, and then the compat tests would still need to pass? (this is likely work for another PR)

Johennes:
> Doesn't this already get run as part of the iOS Compat tests?

This is slightly different actually. The `Package.swift` from above would only compile the xcframework and the Swift files. The compat tests, in turn, generate the React Native bindings and a test app using them which is then compiled.

We could in principle just add the Swift files to the podspec and then they would get compiled together with the compat tests, like you say. However, I think that would only be a vehicle for tests. A consumer would most likely not want to use the Swift bindings through the pod because the pod pulls in the React Native dependencies. The prime use case for the Swift bindings is probably for something like notification service extensions and these have memory and API limitations.

Then again, having the Swift files in the pod doesn't hurt much in the context where you use the pod (React Native / the main app). So we might just do it anyway as long as the people can bundle the Swift files into their extensions separately.

jhugman:
>  A consumer would most likely not want to use the Swift bindings through the pod because the pod pulls in the React Native dependencies.

Oh, I think you're right: and we're not looking to support users who are using swift but not React Native.

> Then again, having the Swift files in the pod doesn't hurt much in the context where you do use the pod (React Native / the main app). So we might just do it anyway as long as people can bundle the Swift files into their extensions separately.

> as long as people can bundle the Swift files into their extensions separately

I'm not sure I understand this bit. Do React Native developers have to do anything extra if there is Swift involved? If so, then is it automatable? e.g. add *.h and *.swift to the podspec?

Johennes:
> > as long as people can bundle the Swift files into their extensions separately
>
> I'm not sure I understand this bit. Do React Native developers have to do anything extra if there is Swift involved? If so, then is it automatable? e.g. add *.h and *.swift to the podspec?

Sorry, I've been speaking sloppily. We can add the Swift files to the podspec but that will only make them available in the main app. When building iOS apps, you add files and libraries to so called "targets". The main app is a target but app extensions are targets of their own. This is because extensions run as separate processes within the main app.

When building an iOS app with, say, the UBRN-generated https://github.com/unomed-dev/react-native-matrix-sdk, React Native will automatically add the pod (which contains the xcframework) to the main app target and that lets you use the TS bindings.

If you now want to add a notification service extension to the app that uses the xcframework through the Swift bindings, you'll need to manually add both of these to the extension's target. This essentially just requires clicking a few buttons in Xcode and pointing it to the paths on disk.

I think we could possibly semi-automate it so that you only need to add some lines in your app's `Podfile`. That would probably require generating a second podspec because the current one includes React Native libraries which in my testing don't compile in app extension targets due to their limited API. We would have to test whether having the xcframework in two pods duplicates it in the final app binary. If so, that could be a problem for larger libraries.

For a first shot, adding the Swift files to the current podspec and letting people manually add them to their extensions feels ok to me. The only downside is that people will compile and link the Swift files twice (once for the main app and once for the extension). But that's probably a rather small inconvenience only.

Johennes:
> For a first shot, adding the Swift files to the current podspec and letting people manually add them to their extensions feels ok to me. The only downside is that people will compile and link the Swift files twice (once for the main app and once for the extension). But that's probably a rather small inconvenience only.

Have made this change with my latest commit. Hoping for the CI to pass now. 🤞

Johennes:
I think the remaining build errors should be fixed by https://github.com/jhugman/uniffi-bindgen-react-native/pull/217.

jhugman (approved):
This is good stuff. Thank you. 

I have no landed #217 to fix the build for this PR.

PR #221, opened 2025-02-26
Enable async-calls fixture for WASM!

jhugman:
This PR splits up the futures fixture into two new fixtures:

- the `async-calls` fixture. This is enabled for wasm and jsi.
- the `async-callbacks` fixture. This is enabled for jsi only.

The `futures` fixture is left in place, as a parity with uniffi-rs, but will, at most, be not updated.

The PR also does some minor refactoring to enable futures to work.


zzorba (approved):
LGTM

PR #222, opened 2025-02-27
Fixup clippy failures

jhugman:
This small PR does two things:

- Fixes a couple of long standing clippy failures.
- Ensures that clippy failures break the build in CI.


zzorba (approved):
LGTM

PR #226, opened 2025-03-05
Ext-types fixture working with WASM!

jhugman:

This small PR adds multi-crate support.

The target crate can generate C-ABI anywhere in its module structure, even within sub-crates. In the generated crate, we are reproducing the C-ABI per module.

The linking happens at the top level "entrypoint" `lib.rs` file, which `use`s the crate's exported library.

Most of this PR is just feeding the Crate Metadata into the entrypoint generation code.


crates/ubrn_bindgen/src/bindings/gen_rust/mod.rs
```diff
@@ -182,15 +188,12 @@ impl<'a> ComponentTemplate<'a> {
...
-    fn prelude(&self, ci: &ComponentInterface) -> TokenStream {
+    fn prelude(&self, _ci: &ComponentInterface) -> TokenStream {
         let runtime_module_ident = self.flavor.runtime_module();
         let runtime_ident = self.runtime_ident();
-        let crate_ident = ident(ci.crate_name());
         let uniffi_ident = self.uniffi_ident();
         quote! {
             use #runtime_module_ident::{self as #runtime_ident, uniffi as #uniffi_ident, IntoJs, IntoRust};
-            #[allow(unused_imports)]
-            use #crate_ident;
```
jhugman, line 193: This was working for single crate fixtures, because there was one `ComponentInterface` and only one crate.

crates/ubrn_bindgen/src/bindings/gen_rust/mod.rs
```diff
@@ -59,9 +63,11 @@ pub(crate) fn generate_entrypoint(modules: &[ModuleMetadata]) -> Result<String>
     formatted(src, true)
 }
 
-fn entrypoint_prelude() -> TokenStream {
+fn entrypoint_prelude(crate_: &CrateMetadata) -> TokenStream {
+    let crate_ident = ident(crate_.library_name());
     quote! {
-        // Generated by uniffi-bindgen-react-native. Now with generating rust
+        #[allow(unused_imports)]
+        use #crate_ident;
```
jhugman, line 73: This now imports the target crate that depends on the other crates.

It's an unused import, because the rust compiler doesn't need it, but the linker does.

jhugman (commented):
Self-review.

PR #230, opened 2025-03-12
Move caching of future-continuation-callback for better hot-reloading

### LINKED ISSUE #224: async call hangs indefinitely after restarting the metro server
Hi there !

react-native v0.76.6

jhugman:
Fixes #224 

This PR eliminates a `static` local variable which was holding on to the `RustFutureContinuationCallback` between cycling of the `jsi::Runtime`.

This was there as a performance optimization, to prevent pointless calls to the `makeCallbackFunction` function.

Now, `makeCallbackFunction` is called each time the Future is polled. To prevent `makeCallbackFunction` doing a lot of duplicated work, we also check that the lambda doesn't exist before making a new one. This lambda is already being cleared by the module destructor.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 [//]: # (## ⚠️ Breaking Changes)
 [//]: # (**Full Changelog**: https://github.com/jhugman/uniffi-bindgen-react-native/compare/{{previous}}...{{current}})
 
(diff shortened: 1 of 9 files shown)


zzorba (approved):
LGTM, thank you for the quick work on this James

jhugman:
Closed in favour of backporting #232

PR #232, opened 2025-03-14
Hot-reloading: ensure jsi::Runtime can be destroyed and remade.

### LINKED ISSUE #224: async call hangs indefinitely after restarting the metro server

jhugman:

Fixes #224.

This PR eliminates static local variables which were holding on to:

- the `RustFutureContinuationCallback` between cycling of the `jsi::Runtime`.
- callback interface vtables which needed to be stored while a pointer was given to Rust.

The test-runner now runs every test twice with two different `jsi::Runtimes` but the same Rust, ensuring that every structure works across a reload of the Javascript (i.e. simulating a hot-reload).

This PR should be backported to `releases/v0.28.x`, currently: #230.

### DIFF
--- a/typescript/src/handle-map.ts
+++ b/typescript/src/handle-map.ts
@@ -20,17 +20,38 @@ export class UniffiHandleMap<T> {
   get(handle: UniffiHandle): T {
     const obj = this.map.get(handle);
     if (obj === undefined) {
+      // Rust is holding a handle which is no longer in the handle map, either
+      // because this is a different handle map to the one it was inserted in,
+      // or that the handle has already been removed from the handlemap it was
+      // originally in.
+      //
+      // This is because of either:
+      //   a) the Typescript has changed state without resetting a callback
+      //      interface, i.e. a hot reload.
+      //   b) a bug in uniffi-bindgen-react-native.
+      //
+      // If this error is thrown when the app is in the wild, i.e. outside of a
+      // development, i.e. not a hot reload, then please file a bug with
+      // uniffi-bindgen-react-native.
+      //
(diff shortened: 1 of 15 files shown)


zzorba (commented):
With this branch (and not the other one), I'm getting an error with hot reload

Not sure if they need to be run together, or this is a replacement for the other ones.

`Error: The object in the handle map has been dropped already`
Screenshot 2025-03-14 at 2 24 40 PM

typescript/src/handle-map.ts
```diff
@@ -20,17 +20,38 @@ export class UniffiHandleMap<T> {
...
+      // uniffi-bindgen-react-native.
+      //
+      // Otherwise, this error is not recoverable, and a cold reload is
+      // necessary.
+      //
+      // If the error is not intermittent, i.e. happening every reload, then
+      // you should probably consider changing the Rust library to not hold
+      // on to callback interfaces and foreign trait instances across reloads,
+      // e.g. creating app or page lifecycle API, or replacing rather than
+      // appending listeners.
```
jhugman, line 44: For a future change, I would like to test if making the initial handle _different_ each run might be a good idea. 

```typescript
export const defaultUniffiHandle = BigInt(Date.now());
```

In the general case, it would be correct, but in the special case of the mostly small changes made to typescript, then it may just become annoying. 

I think we need to probe how hot-reload works, and how often a stale handle is detected in the correct case before changing this.

jhugman (commented):
Self-review.

zzorba (approved):
LGTM

PR #233, opened 2025-03-17
[Backport #232] Hot-reloading: ensure jsi::Runtime can be destroyed and remade

### LINKED ISSUE #224: async call hangs indefinitely after restarting the metro server

jhugman:
According to [The Big O of Code
Reviews](https://www.egorand.dev/the-big-o-of-code-reviews/), this is a O(_n_) change.

Fixes #224.

This PR eliminates static local variables which were holding on to:

- the `RustFutureContinuationCallback` between cycling of the `jsi::Runtime`.
- callback interface vtables which needed to be stored while a pointer was given to Rust.

The test-runner now runs every test twice with two different `jsi::Runtimes` but the same Rust, ensuring that every structure works across a reload of the Javascript (i.e. simulating a hot-reload).

This PR should be backported to `releases/v0.28.x`, currently: #230.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,10 +4,15 @@
 [//]: # (## ⚠️ Breaking Changes)
 [//]: # (**Full Changelog**: https://github.com/jhugman/uniffi-bindgen-react-native/compare/{{previous}}...{{current}})
 
+## 🦊 What's Changed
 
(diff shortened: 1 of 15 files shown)


zzorba (approved):
LGTM

PR #237, opened 2025-03-19
Async callbacks now work in WASM!

jhugman:
This PR does a number of things in order to make the `async-callbacks` fixture work:

- moved the cross `TimerFuture` from `async-calls` fixture in to a `ubrn_testing` crate.
- we make a new fixture, called `async-traits` which was part of `futures`, then moved to `async-callbacks`. It is _not_ a `with_foreign` trait, so didn't seem to be in the right fixture.
- we separate out the various `FfiCallbackFunction` and `FfiStruct`, which uniffi-rs produces in abundance.
- add a Function Literal struct which is passed to Javascript.
  - fix up the template typescript to use this struct `call` method or the `Function#call` method.
- removed the use `RustCallStatus` for async-callbacks, so it is only used in synchronous callback interfaces now.


zzorba (approved):
LGTM, great progress

PR #238, opened 2025-03-20
Fixtures chronological and gc-callbacks-crasher now work with WASM!

jhugman:
This is a mop up PR, which could only really be done once the others have been done.

### `gc-callbacks-crasher`
This enables `gc-callbacks-crasher`, which was really figuring out how to build the crate.

### `chronological`
`chronological` is a little trickier: `wasm32-unknown-unknown` doesn't have access to the clock, so we could build out a web_sys implementation; however, enough of the test worked—i.e. testing the serialization and deserialization of Instant and Duration without re-implementing `Date.now()`.

I _think_ timestamp and duration were implemented before custom types were implemented, which would be the way I would recommend implement time and dates now.

### DIFF
--- a/fixtures/chronological/tests/bindings/.supported-flavors.txt
+++ b/fixtures/chronological/tests/bindings/.supported-flavors.txt
@@ -1,0 +1,2 @@
+jsi
+wasm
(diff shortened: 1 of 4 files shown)


zzorba (approved):
LGTM

PR #239, opened 2025-03-26
Add to "who is using" section of readme

trevoranderson:
Hello ubrn maintainers!
I have an app in the iOS app store that uses ubrn. All my business logic is in Rust with React Native used for the UI layer. The app is not open source, and offers in-app-purchases, but the core functionality is free to use to show people what is possible using the combination of Rust + React Native to combine the performance of Rust with the UI of React Native in a nontrivial application with synchronous and async UI updates.

This pull request adds to the "who is using" section of the readme with a link to where people can download the app and an endorsement blurb about why ubrn has been good for the project.

Open to feedback: thanks for this library!


zzorba (approved):
LGTM, glad the library is working well!

PR #240, opened 2025-03-29
Fix wrong key name of `manifestPath` in docs

EATSTEAK:
This pull request resolves a documentation issue that causes confusion when setting `manifestPath` in `ubrn.config.yaml`.


jhugman (approved):
I'm embarrassed at how this happened. Thank you so much for finding it, and raising a PR!

PR #242, opened 2025-04-08
Bump bob version

Johennes:
This bumps the bob and rn versions to that the non-latest CI jobs pass again.


jhugman (approved):
This is excellent, @Johennes . Thanks.

I have kicked off the Latest jobs just on the off chance that upstream fixed the problem, but please don't wait for that to finish before landing.

PR #252, opened 2025-04-29
Mark fields public on FfiConverterNumber

zzorba:


jhugman (approved):
Making these fields public is fine as `FfiConverterNumber` is not exposed to the user.

PR #253, opened 2025-05-10
Add `ubrn build web --and-generate` command

jhugman:
This adds extra functionality to the `ubrn build web`  command; the following is done automatically:

- Builds the crate for the current architecture, with the set of features taken from the config file.
- Generates the wasm bindings crate, with the correct set of features enabled.
- Generates the typescript bindings
- Runs `cargo build --target wasm32-unknown-unknown`, with the correct profile, taken from the command line.
- Runs `wasm-bindgen` on the resulting wasm bundle.

More command line options are added:

- `--no-generate`: opts out of all except the initial `cargo build`.
- `--no-wasm-pack`: opts out of  the build for the wasm32-unknown-unknown and wasm-bindgen, which does the same as `wasm-pack build`.
- `--target`: changes the target passed to wasm-bindgen

Extras added to the config file:

```yaml
web:
  wasmCrateName`: # defaults to the name taken from `package.json`
  manifestPath: # defaults to `rust_modules/wasm/Cargo.toml`
  workspace: # defaults to false. If false, then assume a workspace `Cargo.toml` is not present.
  features: # defaults to []. The list of features used to build the target crate.
  targets: # list of wasm target triples. Defaults to ["wasm32-unknown-unknown"]
  cargoExtras: # passed to the wasm32 `cargo build`
  target: # default to "web", passed to `wasm-bindgen`
```

Tests are missing at the moment, awaiting a better way of doing integration testing.


crates/ubrn_cli/src/commands/building.rs
```diff
@@ -58,6 +64,7 @@ impl BuildCmd {
         let mut files = match self {
             Self::Android(a) => a.build()?,
             Self::Ios(a) => a.build()?,
+            Self::Web(a) => a.build()?,
```
zzorba, line 67: does this code need wasm cfg/clap stuff as well, won't it be undefined if the wasm feature is missing?

crates/ubrn_cli/templates/wasm/index.web.ts
```diff
@@ -0,0 +1,41 @@
...
+import wasmPath from "./{{ bindings }}/wasm-bindgen/index_bg.wasm";
+
+export async function uniffiInitAsync() {
+  const response = await fetch(wasmPath);
+  const module = await response.arrayBuffer();
+
+  initSync({ module });
+
+  // Initialize the generated bindings: mostly checksums, but also callbacks.
+  // - the boolean flag ensures this loads exactly once, even if the JS code
```
zzorba, line 27: is this comment still true?

zzorba (approved):
LGTM, couple possible nits

crates/ubrn_cli/templates/wasm/index.web.ts
jhugman: yes.

PR #260, opened 2025-05-28
Fixup build for compat with react-native and create-react-native-library

jhugman:
1. Detect versions of CRNL, and use the command line accordingly.
2. Upgrade the versions of react-native and CRNL used in the compat jobs (reducing the number of jobs)
3. Reducing the number of jobs further by changing the schedule for non-latest compat jobs to once per week, instead of once per day.

image


.github/workflows/compat-android.yml
```diff
@@ -2,7 +2,7 @@ name: Compat / Android
 
 on:
   schedule:
-    - cron: "0 0 * * *" # Once a day at midnight
+    - cron: "0 0 * * 1" # Once a week at midnight on Monday
```
Johennes, line 5: I had actually assumed we would fall into GitHub's free tier because the project is open source. Thanks for actually monitoring the cost. 😳

jhugman: It does. However, I'd like to reduce energy use even if I'm not paying for it. 💚 🌍 🌳

Johennes: Oh, even better call then! 👍

Johennes (approved):
LGTM 👍

PR #261, opened 2025-05-29
Change Windows path separators in CMakeLists.txt

### LINKED ISSUE #254: Windows - Build Error due to path separators in cmakelists.txt for android
Transferred from https://github.com/unomed-dev/react-native-matrix-sdk/issues/26. This is a path separator issue when building for Android on Windows.

The ubrn version used in react-native-matrix-sdk at the time of reporting was 0.28.3-1.

> I was getting failed builds on windows trying to build with expo for android. I don't have the error messaging now but the fix i put into place was as follows
>
> in node_modules/@unomed/react-native-matrix-sdk/android/CMakeLists.txt
>
> ```cmake
>    # Convert Windows backslashes to forward slashes for regex compatibility
>    string(REPLACE "\\" "/" UNIFFI_BINDGEN_PATH ${UNIFFI_BINDGEN_PATH})
>```

jhugman:
Fixes #254 .

This copies the fix provided by @skyrover7 in https://github.com/unomed-dev/react-native-matrix-sdk/issues/26.

### DIFF
--- a/crates/ubrn_cli/templates/jsi/android/CMakeLists.txt
+++ b/crates/ubrn_cli/templates/jsi/android/CMakeLists.txt
@@ -17,10 +17,9 @@ execute_process(
     OUTPUT_VARIABLE UNIFFI_BINDGEN_PATH
     OUTPUT_STRIP_TRAILING_WHITESPACE
 )
-string(REGEX
-  REPLACE "/package\\.json$" ""
-  UNIFFI_BINDGEN_PATH ${UNIFFI_BINDGEN_PATH}
-)
+# Get the directory; get_filename_component and cmake_path will normalize
+# paths with Windows path separators.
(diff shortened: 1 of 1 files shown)


skyrover7:
@jhugman note that in the original issue, once I upgraded to 0.50.0, I had to update the package.json line as well, I provided that in the comments of that issue

jhugman:
> @jhugman note that in the original issue, once I upgraded to 0.50.0, I had to update the package.json line as well, I provided that in the comments of that issue

Ah, right you are. 

I took a step back and got cmake to do the work, rather than doing our own string munging.

Johennes (approved):
I cannot test this because I don't have a Windows machine but codewise it looks good. Bonus points for letting cmake wrangle the paths instead of using string manipulation. 👌

PR #263, opened 2025-06-03
Add Cargo.toml extras

jhugman:
- Added a manifest patch file option, `manifestPatchFile`, to `ubrn.config.yaml`. If present, it will be loaded and used to patch over the generated `Cargo.toml`.
- Added default features boolean, `defaultFeatures` to `ubrn.config.yaml`. If present, it works in the same way as `features`: used to build the rlib, then put into the generated `Cargo.toml`.


zzorba (approved):
LGTM

PR #283, opened 2025-07-17
Fix: generating native Kotlin bindings

exploIF:
## Why do we need this?
Currently generating native Kotlin bindings does not work, even that files are created, we are not able to use it, because of couple missing pieces. This PR should add missing logic and improve docs.

## Changes
- added missing implementation in generated `build.gradle`. Logic here is added conditionally, as this is needed only when building native bindings
- added `proguard-rules.pro` - base implementation required to make `JNA` work properly in release builds with proguard turned on
- added missing docs - mostly describing how to configure project and use lib to generate native Kotlin bindings. Most important parts are proguard rules configuration and `cdylib_name` in `uniffi.toml` (which is used by JNA to properly load .so file)


zzorba (approved):
LGTM, thank you for this change.

docs/src/reference/commandline.md
```diff
@@ -128,6 +128,23 @@ crate-type = ["staticlib"]
...
+```toml
+[bindings.kotlin]
+cdylib_name = "<your_lib_name>"
+```
+* if you are using proguard, you will need to add proper rules to the `proguard-rules.pro` file, like so:
+```
+-dontwarn java.awt.*
+-keep class com.sun.jna.* { *; }
+-keepclassmembers class * extends com.sun.jna.* { public *; }
+-keep class <crate_package_name>.** { *; }
```
Johennes, line 144: If I understand the code correctly, ubrn itself already generates the first three lines. If the last line is also required, could ubrn generate it, too?

docs/src/reference/commandline.md
```diff
@@ -369,8 +386,14 @@ Arguments:
   [NAMESPACES]...  The namespaces that are generated by `generate jsi bindings`
 
 Options:
-      --config <CONFIG>  The configuration file for this build
-  -h, --help             Print help
+      --config <CONFIG>  
+          The configuration file for this build
+
+      --native-bindings 
+          This will add required implementations to the generated `build.gradle` file.
```
Johennes, line 393: As written, it might be easy to misunderstand that this is only required when also generating native bindings.
 
```suggestion
          This will add implementations required for native Android bindings to the generated `build.gradle` file.
```

Johennes (approved):
Looks great, thank you! Just one nit and one question.

docs/src/reference/commandline.md
exploIF: I was thinking about that, however I couldn't find a reasonable way of retrieving package name of each crate, as it can be overridden by `uniffi.toml` or generated automatically.

Btw, after your comment I've slightly modified docs, as there is no need to implement first three lines in app's proguard rules, if they are already specified by library itself.

PR #287, opened 2025-07-22
feat: optional record fields

joe-p:
Currently `Optional` types in records are required fields in the generated typescript. This PR replaces the `undefined` union with a `?` to indicate that the field is optional. This results in more idiomatic TypeScript. 

The diff of the generated TypeScript can be seen in this commit: https://github.com/algorandfoundation/algokit-core/commit/30d9488366c67a55e39fbca1deaf99ff5c7e8982

### DIFF
--- a/crates/ubrn_bindgen/src/bindings/gen_typescript/filters.rs
+++ b/crates/ubrn_bindgen/src/bindings/gen_typescript/filters.rs
@@ -203,3 +203,19 @@ pub fn docstring(docstring: &str, spaces: &i32) -> Result<String, askama::Error>
     let spaces = usize::try_from(*spaces).unwrap_or_default();
     Ok(textwrap::indent(&wrapped, &" ".repeat(spaces)))
 }
+
+pub(super) fn is_optional_type(
+    as_type: &impl AsType,
+    types: &TypeRenderer,
+) -> Result<bool, askama::Error> {
+    let type_ = types.as_type(as_type);
+    Ok(matches!(type_, Type::Optional { .. }))
+}
+
+pub(super) fn type_name_without_undefined(
+    as_type: &impl AsType,
+    types: &TypeRenderer,
+) -> Result<String, askama::Error> {
+    let og_type_name = type_name(as_type, types)?;
+    Ok(og_type_name.replace(" | undefined", ""))
+}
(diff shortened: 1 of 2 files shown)


joe-p:
> @joe-p does this better align with what you've seen in other uniffi bindings?

My main experience so far is with the Python bindings, which do use `typing.Optional` meaning they can be omitted, which aligns with that this PR does. For example:

```py
class Transaction:
    transaction_type: "TransactionType"
    """
    The type of transaction
    """

    sender: "str"
    """
    The sender of the transaction
    """

    fee: "typing.Optional[int]"
    """
    Optional transaction fee in microALGO.

    If not set, the fee will be interpreted as 0 by the network.
    """

    first_valid: "int"
    last_valid: "int"
    genesis_hash: "typing.Optional[ByteBuf]"
    genesis_id: "typing.Optional[str]"
    note: "typing.Optional[ByteBuf]"
    rekey_to: "typing.Optional[str]"
    lease: "typing.Optional[ByteBuf]"
    group: "typing.Optional[ByteBuf]"
    payment: "typing.Optional[PaymentTransactionFields]"
    asset_transfer: "typing.Optional[AssetTransferTransactionFields]"
    asset_config: "typing.Optional[AssetConfigTransactionFields]"
    application_call: "typing.Optional[ApplicationCallTransactionFields]"
    key_registration: "typing.Optional[KeyRegistrationTransactionFields]"
    asset_freeze: "typing.Optional[AssetFreezeTransactionFields]"                  
```

>  though perhaps it should be configurable

It could be, but this PR isn't a breaking change and semantics are preserved so I don't think it needs to be. Fwiw I've never seen an `undefined` union in a type/interface definition in the wild, so I think `?` is more idiomatic. 

Also I see CI is failing. Presumably I need to regenerate some generated bindings somewhere, but will further investigate today or tomorrow.

jhugman:
> Also I see CI is failing. Presumably I need to regenerate some generated bindings somewhere, but will further investigate today or tomorrow.

I think the tests are failing because `undefined` is being used in the tests themselves instead of omitting the property or argument. 

On the whole, I don't have any feeling one way or another about this but have these cautionary tales / silly questions to tell:

- we have [more and more users](https://www.npmjs.com/package/uniffi-bindgen-react-native?activeTab=dependents), so surprising non-optional breakages should be kept close zero. I think I'd agree with @zzorba in asking for this be triggered on a property in `uniffi.toml`.
- how does this interact with default values?
- when adding default values to python, we found that python supports default arguments only in the final arguments in a list. I'm not sure how the parsing works in Typescript, or even if we should mirror Python's behavior if Typescript allows it: in a general case, I don't think I'd like to be able write Rust which could be called from Typescript, but not from Python.

Issue #295, opened 2025-09-12
`mHybridData` field access causes crash on React Native 0.80+ and expo 54 - Android compatibility issue


Leeeon233:
Thank you for providing such an excellent library! We're using `uniffi-bindgen-react-native` for [Loro](https://github.com/loro-dev/loro)'s React Native bindings, and our community members have discovered a compatibility issue.

When running on React Native 0.80 and above or Expo 54, Android applications crash immediately on startup with the following error:
```
09-11 15:13:56.289 27682 27682 F DEBUG   : signal 6 (SIGABRT), code -1 (SI_QUEUE), fault addr --------
09-11 15:13:56.289 27682 27682 F DEBUG   : Abort message: 'JNI DETECTED ERROR IN APPLICATION: JNI GetObjectField called with pending exception java.lang.NoSuchFieldError: no "Lcom/facebook/jni/HybridData;" field "mHybridData" in class "Lcom/facebook/react/turbomodule/core/CallInvokerHolderImpl;" or its superclasses
09-11 15:13:56.289 27682 27682 F DEBUG   :   at boolean com.loro.lororeactnative.LoroReactNativeModule.nativeInstallRustCrate(long, com.facebook.react.turbomodule.core.interfaces.CallInvokerHolder) (LoroReactNativeModule.kt:-2)
09-11 15:13:56.289 27682 27682 F DEBUG   :   at boolean com.loro.lororeactnative.LoroReactNativeModule.installRustCrate() (LoroReactNativeModule.kt:24)
09-11 15:13:56.289 27682 27682 F DEBUG   :   at void com.facebook.jni.NativeRunnable.run() (NativeRunnable.java:-2)
```

This appears to seem to be related to the `mHybridData` [field access](https://github.com/jhugman/uniffi-bindgen-react-native/blob/0fb3841b13bcb1bcd8d98f31c6dda8334564d916/crates/ubrn_cli/templates/jsi/android/cpp-adapter.cpp#L41), Realm.js encountered the exact same issue and has already implemented a fix. Their investigation and solution might provide valuable insights for addressing this compatibility problem:

Related Issue: https://github.com/realm/realm-js/issues/7011

Thank you for your time and consideration!

PR #298, opened 2025-09-15
fix: use public fbjni API

exploIF:
## Explanation
Current approach worked fine until react-native 0.80. The issue is that it's using private API, which has been changed in the react-native 0.80. With new approach we are using public API. This fixes runtime issues on Android in react-native 0.80 and 0.81.

## Testing strategy
I've tested that Android builds and runs fine on react-native 0.79 and 0.81


exploIF:
Fixes https://github.com/jhugman/uniffi-bindgen-react-native/issues/295

zzorba:
Yeah, this test is currently failing at main, I think updating the reference copies is correct, but I also don't know how we would do that.

n8mgr:
Do you have a timeline for release of this patch? Hoping to avoid downgrading react native 😅.

PR #300, opened 2025-09-15
Fix CI cargo version by pinning it to 1.88.0

zzorba:


zzorba:
@jhugman this fixes one of the two failing tests @main, but not the bob one.  Still worth reviewing + merging I believe.

.github/workflows/ci.yml
```diff
@@ -16,6 +16,13 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@1.88.0
+
+      - name: Install Rust Fmt and Clippy
+        shell: bash
+        run: rustup component add rustfmt clippy
```
Johennes, line 24: If we want to also enforce this outside the CI, we could use the `rust-toolchain.toml` file. If I understand the [documentation](https://rust-lang.github.io/rustup/overrides.html) correctly, `rustup` should then automatically install / pick the specified version when commands are issued.

Johennes: I think it's just a file at the repo level and rustup should use it automatically. I've never used it myself so far though. Not a real issue. Was just a thought.

jhugman: I think @Johennes is right; if we don't want to do that, then we can [put it in the manifest](https://doc.rust-lang.org/cargo/reference/rust-version.html).

jhugman (approved):
I think now that @Johennes has fixed the other issues, this should build cleanly. Good work team!

PR #307, opened 2025-09-27
Disable typescript checking and linters in generated typescript files

Marcel-Nordeck:
Without this, eslint, tslint and typescript depending on the configuration will complain about the generated code. This is especially easy to hit with the vitejs default tsconfig setup, which is stricter.

Just as a disclosure, this PR I am doing for my employer Nordeck, and it's customer as part of the https://gitlab.opencode.de/fitko/matrix-g2x project.


Johennes (approved):
Looks sane. CI failures are the same as on `main` (and we should fix them at some point).

PR #310, opened 2025-09-30
Stop checking for jsRootDir & friends (?)

Johennes:
Once https://github.com/jhugman/uniffi-bindgen-react-native/pull/309 has landed, this should "fix" the Compat tests. I'm not sure if this is a good fix though. Apparently CRNL doesn't use `jsRootDir` anymore when generating libraries which means we cannot really check for it remaining untouched anymore. As a corollary, this may mean that we shouldn't use `jsRootDir` ourselves either when generating `build.gradle`. I'm not aware of this causing any issues, however, so just dropping the check might be ok-ish for now?

Edit: Apparently the `libraryName` and `codegenJavaPackageName` properties cause the same problem making this even less of a great fix. 🫤

### DIFF
--- a/scripts/test-turbo-modules.sh
+++ b/scripts/test-turbo-modules.sh
@@ -308,7 +308,6 @@ check_lines() {
 
   check_line_unchanged "./android/CMakeLists.txt" "^project"
   check_line_unchanged "./android/CMakeLists.txt" "^add_library.*SHARED"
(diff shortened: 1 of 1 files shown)


jhugman (approved):
Thank you for doing this, @Johennes . It _really_ helps even though it might not seem like it.

Issue #306, opened 2025-09-26
`async` and `public` are swapped in the typescript bindings


antonWetzel:
### Usage

- Use Rust with uniffi for a website (no react or other frameworks)

### Problem

- The generated bindings contain `async public method(...)`, which is not valid typescript.

### Ideas for Solutions

- remove the `public` from the template (I think typescript uses public by default)
- change the keyword order

#### Rust

```rust
uniffi::setup_scaffolding!();

#[derive(Debug, uniffi::Object)]
pub struct Factory {}

#[uniffi::export]
impl Factory {
    #[uniffi::constructor]
    pub fn new() -> Self {
        Self {}
    }

    pub async fn create_product(&self) -> Product {
        Product {}
    }
}

#[derive(Debug, uniffi::Object)]
pub struct Product {}
```

#### Generated Typescript

```ts
// autogenerated
export class Factory extends UniffiAbstractObject implements FactoryInterface {
    
    ...

    async public createProduct(asyncOpts_?: { signal: AbortSignal }): Promise<ProductInterface> {
       ...
    }

    ...
}
```

#### Repository
- https://github.com/antonWetzel/uniffi-remote-async/tree/uniffi-web

PR #319, opened 2025-10-21
Remove public access modifier for the TypeScript bindings

antonWetzel:
Fixes https://github.com/jhugman/uniffi-bindgen-react-native/issues/306

This removes the public access modifier for method declarations and the error tag.
Because public is the default modifier, this should not change behavior.

This is only tested with a example project, not the project tests, see notes for details.

## Note for tests
- Platform: `macOS`
- Installed for bootstrap `cmake` and `ninja`
- Error during bootstrap `ld: library 'hermes' not found`
- Same error for `./scripts/run-tests.sh`


EthanShoeDev:
I have never seen pr screen missing the merge button... weird. Please merge this!

Issue #324, opened 2025-11-14
package.json and tsconfig.json `module` values conflict


spacebear21:
Disclaimer: I am not a JS expert and used Claude liberally to troubleshoot and explain things to me.

I am using this library to generate WASM + JS bindings that can run in Node.js (https://github.com/payjoin/rust-payjoin/pull/1190). I've got it mostly working by just setting `target: nodejs` in the `ubrn.config.yaml`. However, there was an issue with module imports resulting in the error below when executing the generated code:

```
❯ npm run test

> payjoin@0.1.0 test
> node test.js

file:///Users/spacebear/Projects/rust-payjoin/payjoin-ffi/javascript/dist/generated/bitcoin.js:7
import { AbstractFfiConverterByteArray, FfiConverterArray, FfiConverterArrayBuffer, FfiConverterBool, FfiConverterFloat64, FfiConverterInt32, FfiConverterObject, FfiConverterUInt32, FfiConverterUInt64, FfiConverterUInt8, UniffiAbstractObject, UniffiError, UniffiInternalError, UniffiRustCaller, destructorGuardSymbol, pointerLiteralSymbol, uniffiCreateFfiConverterString, uniffiCreateRecord, uniffiTypeNameSymbol, } from "uniffi-bindgen-react-native";
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
SyntaxError: The requested module 'uniffi-bindgen-react-native' does not provide an export named 'AbstractFfiConverterByteArray'
    at #_instantiate (node:internal/modules/esm/module_job:254:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:369:5)
    at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:691:26)
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:101:5)

Node.js v24.11.0
```

This seems to be caused by a conflict between package.json which specifies `"type": "module"`, and tsconfig.json which sets `"module": "commonjs"`. The package declares itself as ESM but compiles to CommonJS.

I found [a note](https://www.typescriptlang.org/docs/handbook/modules/theory.html#the-module-output-format) in the TS documentation which states:

[...]

PR #325, opened 2025-11-14
Change `module` compiler option to `nodenext`

spacebear21:
See #324 for background.

The TS
[documentation](https://www.typescriptlang.org/docs/handbook/modules/theory.html#the-module-output-format) states that:

"Node.js’s rules for module format detection and interoperability make it incorrect to specify module as esnext or commonjs for projects that run in Node.js, even if all files emitted by tsc are ESM or CJS, respectively. The only correct module settings for projects that intend to run in Node.js are node16 and nodenext. While the emitted JavaScript for an all-ESM Node.js project might look identical between compilations using esnext and nodenext, the type checking can differ. See the reference section on nodenext for more details."

This change makes the project runnable in Node.js by changing the module compiler option to "nodenext", and fixing the imports accordingly.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "target": "es2022",                                  /* Set the JavaScript language version for emitted JavaScript and include compatible library declarations. */
-    "module": "commonjs", 
+    "module": "nodenext", 
     "declaration": true,
     "outDir": "./typescript/dist",                                   /* Specify what module code is generated. */
     "paths": {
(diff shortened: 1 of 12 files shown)


spacebear21:
Reopened and rebased because this fix is still needed as evidenced by discussions in #370 and #375 

Also have a minimal repro step which gets fixed with this PR:

```
  yarn build
  node --input-type=module \
    -e "import('./typescript/dist/index.js').then(m => console.log(Object.keys(m).length))"
```

spacebear21:
Closing this as obsolete, the project is now runnable in Node.js via typescript/tsconfig.esm.json as of #399 🥳

PR #330, opened 2025-12-02
Add --tools parameter introduced in crnl@0.55.0

Johennes:
This fixes the compat CI failures on `main`.

### DIFF
--- a/scripts/test-turbo-modules.sh
+++ b/scripts/test-turbo-modules.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -e
+set -ex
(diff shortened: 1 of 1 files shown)


scripts/test-turbo-modules.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -e
+set -ex
```
Johennes: This logs shell commands which greatly helps debugging issues.

jhugman (approved):
This will be a relief to see this landed. Thanks @Johennes !

PR #341, opened 2026-02-03
Strict object types: opt out of generating and using typescript interfaces

### LINKED ISSUE #328: Generated TypeScript uses interface instead of class for `uniffi::Object`, reducing type safety
When defining a `uniffi::Object` in Rust, `ubrn build web` generates both a TypeScript interface and a TypeScript class for the Rust struct.
However, whenever this object is used in records or function signatures, the generated code refers to the interface instead of the actual class.

Since the class has strong typing, but the interface is only a weak type, this results in weaker type guarantees for consumers of the generated bindings.

Here a short example, assume the following rust code:

```rust
#[derive(uniffi::Object)]
pub struct MyObject {}

#[derive(uniffi::Record)]
pub struct MyRecord {
    member: Arc<MyObject>,
}

#[uniffi::export]
pub fn my_function(my_object: Arc<MyObject>) -> Arc<MyObject> {
    my_object
}
```

The ubrn generated typescript code looks like this:

```ts
﻿export interface MyObjectInterface {
    
}

export class MyObject extends UniffiAbstractObject implements MyObjectInterface {

    readonly [uniffiTypeNameSymbol] = "MyObject";
    readonly [destructorGuardSymbol]: UniffiRustArcPtr;
    readonly [pointerLiteralSymbol]: UnsafeMutableRawPointer;
    
    ...
}

export type MyRecord = {
    member: MyObjectInterface
}

export function myFunction(myObject: MyObjectInterface): MyObjectInterface {
    return FfiConverterTypeMyObject.lift(uniffiCaller.rustCall(
            /*caller:*/ (callStatus) => {
                return nativeModule().ubrn_uniffi_rust_src_fn_func_my_function(
        FfiConverterTypeMyObject.lower(myObject),
                callStatus);
            },
            /*liftString:*/ FfiConverterString.lift,
    ));
}

```

In both the record and the function signature, the generated bindings use MyObjectInterface instead of the actual MyObject class. This makes the API less type-safe than expected, since the interface allows any structurally compatible value, while the class enforces correct usage.

The generated TypeScript should refer to the class (MyObject) wherever the Rust type is Arc<MyObject>, providing stronger typing guarantees for consumers.

Sample code is provided here: https://github.com/wireapp/ubrn-wasm-demo/tree/issue/interfaces

SimonThormeyer:
Closes #328

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@
 
 - Records and enums now support methods in TypeScript bindings, tracking uniffi 0.31's new value-type methods feature. Flat enums use direct method calls; tagged enums use a factory pattern ([#347](https://github.com/jhugman/uniffi-bindgen-react-native/pull/347)).
 
+- Add support for opting out of generating TypeScript interfaces: `bindings.typescript.strictObjectTypes` in
+  `uniffi.toml` ([#341](https://github.com/jhugman/uniffi-bindgen-react-native/pull/341) by
+  [@SimonThormeyer](https://github.com/SimonThormeyer)).
+
 ## 🦊 What's Changed
 
 - Bump `uniffi-rs` to [0.31.0](https://github.com/mozilla/uniffi-rs/blob/main/CHANGELOG.md).


crates/ubrn_bindgen/src/bindings/gen_typescript/templates/ObjectTemplate.ts
```diff
@@ -26,7 +29,7 @@ private constructor(pointer: UnsafeMutableRawPointer) {
 {%- endmacro %}
 
 {% call ts::docstring(obj, 0) %}
-export class {{ impl_class_name }} extends UniffiAbstractObject implements {{ protocol_name }} {
+export class {{ impl_class_name }} extends UniffiAbstractObject {%- if needs_interface %} implements {{ protocol_name }}{%- endif %} {
 
     readonly [uniffiTypeNameSymbol] = "{{ impl_class_name }}";
     readonly [destructorGuardSymbol]: UniffiRustArcPtr;
```
jhugman: This is the heart of the implementation.

SimonThormeyer:
Hi @jhugman, I updated this branch according to the discussion in the issue. Thanks for considering our feature request and taking another look.

Issue #360, opened 2026-03-23
Proposal: Javascript context specific dynamic FFIs (a.k.a. player)


jhugman:
Ok, slightly clickbait title.

I am looking at adding node support.

There are two ways to do it: 

1. Approach 1: generate crate specific `napi-rs` bindings. The low level FFI (`FFIType` in uniffi parlance) is moved across from napi world to uniffi world, in the same way as we do for `wasm-bindgen`. We have a generated crate which needs to be compiled and linked on every build: we make a single `.node` file, which the Typescript knows how to load.
2. Approach 2:  inspired by [`ffi-rs`](ffi-rs) popular in the node community, and by [the recent uniffi-bindgen-node project](https://github.com/livekit/uniffi-bindgen-node)—hat tip to @1egoman. It's built with napi-rs and `dlopen`. We have a single `uniffi-javascript-player-node` crate generic across all client crates. The generated low-level typescript is used configure the player crate. The player crate is versioned and distributed on its own via npmjs.

`ffi-rs` is not suited for use with uniffi—the threading falls down especially around callback interfaces. This does not mean we can't build our own. [I have built a prototype, which is here][uniffi-napi].

[ffi-rs]: https://www.npmjs.com/package/ffi-rs
[uniffi-napi]:  https://github.com/jhugman/uniffi-napi

The benefits for the latter over the former are quite significant:

- development cycle: 
    1. Approach 1: Typescript and FFI crate is generated,  the FFI crate needs to be compiled, then linked to the client crate OR
    2. Approach 2: Typescript generated, and we're done. 
- distribution:
    1. Approach 1: it's not clear who is responsible for the generation and compilation of the artifacts
    2. Approach 2: "Binary" only packages are possible, where users don't need Rust or other tooling
- easier adding of other backends. We've moved the large hairy JS-Native FFI from needing to be generated to something more generic. This is makes it much more tractable to add more backends: bun and deno are surely to follow.

## So what?

It's the last point that I'm getting excited by: what if we adopt the same player approach for React Native, and WASM?

- FFI can be regenerated quick enough for React Native and Android at the same time. 
     - No recompile of C++ or kotlin or JNI. I don't think anyone needs to have gradlew to re-build the library. 
     - There may be a re-linking for iOS, but I think that will be minimal. But no-one has to have xcodebuild or swiftc or cocoapods
     - Being able to generate for node, bun and deno all at the same time will be a similar story.
[...]

PR #363, opened 2026-03-25
 Migrate the wrapper.ts generation to the new pipeline API

jhugman:
This builds on #359.

This end up being a mostly a moving code around and cleaning up exercise.

The pipeline from uniffi is now topologically sorted, so we can remove that dependency and code.

The pipeline produces a tree of nodes, which gets progressively transformed.

Then the templates get called. This PR removes the layer of render objects which were tied to individual templates or using askma `include`. This is completely replaced by converting these templates into askama macros: this enables the macros to follow the same structure as the transformed tree.

Close #366


crates/ubrn_bindgen/src/bindings/gen_typescript/ffi_module/nodes.rs
```diff
@@ -10,9 +10,43 @@
...
+            .filter_map(|def| match def {
+                FfiDefinitionDecl::Callback(cb) if cb.exported => {
+                    Some(FfiExportedName::Callback(cb.name.clone()))
+                }
+                FfiDefinitionDecl::Struct(s) if s.exported => {
+                    Some(FfiExportedName::Struct(s.name.clone()))
+                }
+                _ => None,
+            })
+            .collect()
```
jhugman, line 48: This drives the type definitions in both WASM and JSI. They moved here when moving out of the `wrapper.ts`; now we make a file for WASM too.

crates/ubrn_bindgen/src/bindings/gen_typescript/ffi_module/builder.rs
```diff
@@ -29,10 +35,25 @@ impl TsFfiModule {
...
+        let has_continuation_callback = definitions.iter().any(|d| {
+            matches!(d,
+                FfiDefinitionDecl::Callback(cb) if cb.name == "UniffiRustFutureContinuationCallback"
+            )
+        });
+        let has_foreign_future = definitions.iter().any(|d| {
+            matches!(d,
+                FfiDefinitionDecl::Struct(s) if s.name == "UniffiForeignFuture"
+            )
+        });
```
jhugman, line 47: This is driving the type check to see if the generated types are the same as the ones we use in `typescript/src`.

I think this will likely go away.

jhugman:
I think given that [mozilla/uniffi-rs#](https://github.com/mozilla/uniffi-rs/issues/2849), and the context of this PR in the #360 proposal, I am willing to accept the risk of churn across versions of uniffi-rs.

Even with the churn the benefits are: 
- much less logic in the templates.
- stability in the templates—since this project creates most of the data structures that go into the templates.

Given the pipeline model is more or less stable, and we have a smalll number of Rust files transforming the IR into nodes we control, if the API changes, then the changes are likely to be limited to a few Rust files, instead of spread across templates and macros.

crates/ubrn_bindgen/src/bindings/gen_typescript/templates/CallBodyMacros.ts
```diff
@@ -0,0 +1,317 @@
+{#- Shared macros for rendering sync and async call bodies from TsCallable IR nodes.
+
+    These macros expect the following variables to be in the template context:
+    - `is_verbose: bool`
+    - `supports_rust_backtrace: bool`
+    - `console_import: &Option<String>`
+-#}
```
jhugman, line 4: This file is functionally very similar to the now deleted `macros.ts` file.

examples/arithmetic-procmacro/uniffi.toml
```diff
@@ -0,0 +1,2 @@
+[bindings.typescript]
+strictTypeChecking = true
```
jhugman, line 2: These are added here instead to ensure tsc checks the generated bindings. The existing bindings were not being typed checked, and failed this. We should close #366 when this lands.

crates/ubrn_fixture_testing/src/lib.rs
```diff
@@ -57,6 +57,17 @@ impl Drop for CleanupFile {
...
+/// Run a command, inheriting stdout/stderr so output is visible.
+pub(crate) fn run_cmd(cmd: &mut Command) {
+    let status = cmd
+        .status()
+        .unwrap_or_else(|e| panic!("failed to launch {:?}: {e}", cmd.get_program()));
+
+    if !status.success() {
+        panic!("{:?} failed (exit status: {})", cmd.get_program(), status);
+    }
+}
```
jhugman, line 69: This let's us verify that the tests are actually running.

crates/ubrn_bindgen/src/bindings/gen_typescript/api_module/nodes.rs
```diff
@@ -0,0 +1,308 @@
+/*
```
jhugman, line 1: This is the best place to start.

crates/ubrn_bindgen/src/bindings/gen_typescript/api_module/mod.rs
```diff
@@ -0,0 +1,568 @@
...
+#[derive(Debug, Eq, Ord, PartialEq, PartialOrd)]
+enum ImportedItem {
+    Type(String),
+    Value(String),
+}
+
+pub(crate) struct ImportAccumulator {
+    imports: BTreeMap<String, BTreeSet<ImportedItem>>,
+    exported_converters: BTreeSet<String>,
+    imported_converters: BTreeMap<(String, String), BTreeSet<String>>,
```
jhugman, line 63: The intended shape of the code is for 

```ts
// imports
import { rustCall, FfiConverterInt32, FfiConverterObject } from 'uniffi-bindgen-react-native';
import nativeModule from './mylib-ffi';
import { type MyCallbackVtable } from './mylib-ffi';

// imported converters
import converters as unffiExtLibFfiConverters, { ExtObject } from './extlib';
const { FfiConverterTypeExtObject } = uniffiExtLibFfiConverters;

// exported convertes
export {
  converters: { FfiConverterTypeMyObject, FfiConverterTypeMyCallback }, 
}
```

crates/ubrn_bindgen/src/bindings/gen_typescript/templates/CallbackInterfaceImpl.ts
```diff
@@ -1,22 +1,17 @@
-{{- self.import_infra_type("UniffiHandle", "handle-map") }}
```
jhugman, line 1: These templates are largely the same except:

- almost all filters are gone: everything is pre-calculated.
- macros not includes: askama macros are like functions, so we can just visit each node in turn.
- templates are rooted to one render struct, not dozens of them.

jhugman: Reviewing is relatively straightforward: each line is a straightforward edit with these three rules.

Once you're through the templates, then the serious deleting of rust code happens.

crates/ubrn_fixture_testing/src/lib.rs
zzorba: "who tests the test rig"

PR #364, opened 2026-03-25
Convert Windows back slash paths to forward slashes

pepperoni505:
#352 already touched on this for CMakeLists.txt, but the issue is still present in all of the other users of `relative_to`. This PR moves the fix from that to the `relative_to` function itself

### DIFF
--- a/crates/ubrn_cli/src/codegen/mod.rs
+++ b/crates/ubrn_cli/src/codegen/mod.rs
@@ -8,6 +8,7 @@ use std::{cell::OnceCell, collections::BTreeMap, rc::Rc};
 use anyhow::Result;
 use askama::DynTemplate;
 use camino::{Utf8Path, Utf8PathBuf};
+use path_slash::PathExt;
 
 use ubrn_bindgen::ModuleMetadata;
 use ubrn_common::{mk_dir, CrateMetadata};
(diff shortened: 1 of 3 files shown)


jhugman:
I'm afraid i haven't a Windows machine to test these changes. 

Thank you for jumping in. 

Generally, I'm happy to fix the problem as far up the stack as it needs to go: is `relative_to` used anywhere else on Windows apart from the `CMakeLists.txt` file?

@DavJCosby can I get your testing fingers on this?

DavJCosby:
I could give it a shot!

Disclaimer: I didn't see anything else break on my specific use case when I ran this so I wouldn't be able to tell you if the change resolved any other bugs. If it ends up breaking anything, though, I'll let you know.

pepperoni505:
> Generally, I'm happy to fix the problem as far up the stack as it needs to go: is `relative_to` used anywhere else on Windows apart from the `CMakeLists.txt` file?

I ran into issues with the generated `index.web.ts`, where paths would look like `import initAsync from './wasm-bindgen\index.js';`. I forget the exact conditions I had that occur under as I am not able to reproduce ATM, but it was 100% an issue

DavJCosby:
Sorry to leave you hanging so long; busy busy week.

Best I can tell, no unintentional side effects of this PR. Much more elegant solution, I approve!

My team decided to make some architectural changes for our project and now we don't really require rust + typescript bindings (😢) so I haven't gotten any more familiar with this project since my PR. Sorry if something breaks in the future for someone else, but again best I can tell this works.

pepperoni505:
@jhugman Can this get merged? 🙏

jhugman (approved):
This appears to be working on Linux, on CI. I haven't been able to test on Windows.

jhugman:
Thanks @pepperoni505 !

PR #367, opened 2026-03-27
Strip \\?\ prefix from canonicalized paths on Windows

pepperoni505:
On Windows, `std::fs::canonicalize` returns paths with a `\\?\` prefix (extended-length path syntax). These paths are not understood by certain tools and cause failures when they're passed to external processes or interpolated into generated files. This was an issue for me when running `ubrn build android`:
```
error: linking with `C:\Users\jackl\.cargo\bin\cargo-ndk` failed: exit code: 1
ld.lld: error: cannot find version script \\\\?\\D:\\Code\\foo\\target\\aarch64-linux-android\\debug\\deps\\rustc6z5PU7\\list
ld.lld: error: cannot open \\\\?\\D:\\Code\\foo\\target\\aarch64-linux-android\\debug\\deps\\rustc6z5PU7\\symbols.o: No such file or directory
ld.lld: error: cannot open \\\\?\\D:\\Code\\foo\\target\\aarch64-linux-android\\debug\\deps\\foo.0dwosz5mz4akfvj7agx7hivx7.082ow32.rcgu.o: No such file or directory
```
(trimmed for easier reading).

I added `dunce`, which was made for this case. Their [README](https://gitlab.com/kornelski/dunce/-/blob/master/README.md) explains the issue in a bit more detail.


jhugman (approved):
This looks good, and runs on Linux too!

PR #365, opened 2026-03-25
Don't prepend `lib` to dlls

pepperoni505:
On Windows, Rust's cdylib output doesn't prefix DLL filenames with `lib`. This issue was noticeable running `ubrn build web`, as the tool searched for `libfoo.dll` instead of `foo.dll`.

PR #368, opened 2026-03-27
Have prettier only format JS and TS files

pepperoni505:
`prettier` was being called to run on every file from where it was invoked. This includes the `rust_modules/wasm/target/` directory, which contains large binary files (1.5gb) that exceed Node's string length limit, causing the build to fail with `Invalid string length`.

This shouldn't be reproducible with the default config, as the folder layout prevents it. My `ubrn.config.yaml` is as follows:
```yaml
rust:
  directory: .
  manifestPath: crates/foo-ffi/Cargo.toml
web:
  ts: generated/web
  manifestPath: generated/web/rust_modules/wasm/Cargo.toml
  entrypoint: generated/web/index.web.ts
bindings:
  cpp: generated/cpp
  ts: generated/src
turboModule:
  cpp: generated/cpp
  ts: generated/src
  entrypoint: generated/src/index.tsx
android:
  directory: generated/android
  packageName: com.bar.foo
ios:
  directory: generated/ios
```

### DIFF
--- a/crates/ubrn_common/src/fmt.rs
+++ b/crates/ubrn_common/src/fmt.rs
@@ -43,7 +43,8 @@ pub fn prettier<P: AsRef<Utf8Path>>(out_dir: P, check_only: bool) -> Result<Opti
         } else {
             cmd.arg("--write");
         }
-        cmd.arg(".").current_dir(out_dir.as_ref());
+        cmd.args(["**/*.js", "**/*.ts"])
+            .current_dir(out_dir.as_ref());
         Some(cmd)
     } else {
         use crate::testing::{is_recording_enabled, record_command};
--- a/crates/ubrn_common/src/fmt.rs
+++ b/crates/ubrn_common/src/fmt.rs
@@ -43,8 +43,13 @@ pub fn prettier<P: AsRef<Utf8Path>>(out_dir: P, check_only: bool) -> Result<Opti
         } else {
             cmd.arg("--write");
         }
-        cmd.args(["**/*.js", "**/*.ts", "**/*.json", "--no-error-on-unmatched-pattern"])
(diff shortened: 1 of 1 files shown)


pepperoni505:
@jhugman Just pushed a fix for the failing tests, which (hopefully, since I can't run them locally) works. Can you re-run workflows?

jhugman:
Re-launched tests. 

> since I can't run them locally

Hmm. Is this a documentation issue at our end, or a local to you issue?

Ooh. [The docs](https://jhugman.github.io/uniffi-bindgen-react-native/contributing/local-development.html#running-tests) could do with a spruce up! Sorry.

Running the jsi and wasm tests for arithmetic example:

```sh
cargo test -p uniffi-example-arithmetic -- jsi::
cargo test -p uniffi-example-arithmetic -- wasm::
```

pepperoni505:
Great, thanks! Fixed the lint error.

> Hmm. Is this a documentation issue at our end, or a local to you issue?

Local issue. I'm on Windows and I'm having a hard time getting the environment setup with building Hermes. I touched on this briefly [here](https://github.com/jhugman/uniffi-bindgen-react-native/pull/365#issuecomment-4156378590) but I realize it's not very helpful 😓. I'm taking a deeper look at this now and I'm getting farther. I'll make a PR when I get it working for Windows

EDIT: What are your thoughts on adding a Windows CI to live alongside the Linux one? I am concerned over time that certain things may drift and cause certain things on Windows to fail in regards to the testing pipeline. The shell scripts under `scripts` will need to be converted to JS or something like that to make them platform agnostic

EDIT 2: Fixed in #373 😉

jhugman (approved):
Great! Thanks for the contribution.

PR #369, opened 2026-03-30
Add uniffi-runtime-napi to runtimes/

jhugman:
## Summary

- Copies [uniffi-napi](https://github.com/jhugman/uniffi-napi) into `runtimes/napi/` and wires it into the Cargo workspace
- Renames crate and npm package to `uniffi-runtime-napi`
- Adds a test that loads the arithmetic example through the napi player
- Adds all 51 runtime tests to CI

## What changed

The napi player is a napi-rs native addon that loads Rust libraries at runtime via `dlopen` and calls their functions through libffi. Unlike the current codegen approach, it needs no generated C++ or Rust per crate. The generated TypeScript describes the function signatures; the player handles the rest.

This is the first step toward the player architecture (#360). The `runtimes/` directory will hold each player — napi, jsi, wasm, and qjs — behind a shared `open()`/`register()`/`FfiType` API.

There is no generation code in this PR, that follows.

Most of the code here is taken up by tests and fixtures. The actual Rust is relatively small. If you're only interested in that, look at `runtimes/napi/src`.

There is quite a lot of `unsafe` code in this player implementation, though: 

- each block is minimized in size and scope.
- I have added `SAFETY` blocks for each of them.
- they fall into the following categories:
  - `libffi`: pointer casts in trampolines, but each CIF is built from the same `FfiTypeDesc` that drives that pointer cast.
  - `napi-rs`: type conversions, reconstructing napi wrapper types from raw handles.
  - function pointer transmutes, which rely on the JS `register` method providing the same types that match the Rust. I think this is going to be hard to detect or circumvent locally, but the mitigations are more about the checksums generated between uniffi-rs and the generated binding.
  - `Send` and `Sync` implementations for the trampolines which are only derefed on the main thread. A runtime check makes sure that we're on the correct thread at entry.

## Test plan

- [ ] `cargo check -p uniffi-runtime-napi` compiles
- [ ] `cargo check -p uniffi-bindgen-react-native` still compiles
- [ ] `cd runtimes/napi && npm install && npm run build:debug && npm test` passes (51 tests)
- [ ] CI job "Integration tests (N-API runtime)" passes


runtimes/napi/fixtures/test_lib/src/lib.rs
```diff
@@ -0,0 +1,906 @@
+use std::ptr;
```
jhugman, line 6: This file simulates what uniffi-rs will produce as part of the Rust scaffolding. We do this without uniffi to start with.

runtimes/napi/fixtures/test_lib/src/lib.rs
```diff
@@ -0,0 +1,906 @@
...
+    status: &mut RustCallStatus,
+) {
+    status.code = 0;
+    std::thread::spawn(move || {
+        std::thread::sleep(std::time::Duration::from_millis(10));
+        cb(handle, value);
+    });
+}
+
+// --- VTable test ---
```
jhugman, line 288: The rest of this file is for callbacks and async calls.

runtimes/napi/fixtures/uniffi-fixture-simple/src/lib.rs
```diff
@@ -0,0 +1,82 @@
+uniffi::setup_scaffolding!();
```
jhugman, line 6: This is an actual uniffi fixture, so we can test against real uniffi generated ABI.
As with each of these tests, they are imported directly from the prototype.

I'd definitely argue that this wasn't the final resting place for these fixtures or tests: having a unified set of fixtures and tests for all player implementions would be beneficial to the project.

runtimes/napi/src/register.rs
```diff
@@ -0,0 +1,610 @@
+//! # The Orchestrator
```
jhugman, line 6: This is a good place to start.

runtimes/napi/src/register.rs
```diff
@@ -0,0 +1,610 @@
...
+//!    buffer contents (converted to `Uint8Array`).
+//! 6. **Marshal the return value** back to JS.
+//!
+//! ## On intentional leaks
+//!
+//! Callback closures and their userdata are intentionally leaked (`Box::into_raw` +
+//! `mem::forget`). The native library may invoke the callback from an arbitrary thread
+//! at an arbitrary time after `call_ffi_function` returns, so we cannot determine the
+//! callback's lifetime statically. This is a deliberate trade-off: a small, bounded
+//! amount of memory is leaked per callback registration in exchange for soundness.
```
jhugman, line 52: "Callback registration" means the machinery to set up each callback interface type, NOT every JS callback causes a leak in memory.

Node doesn't have the concept of unrequiring a module, so we never unregister callbacks either.

runtimes/napi/src/register.rs
```diff
@@ -0,0 +1,610 @@
...
+        free_ptr,
+    })
+}
+
+/// Build a JS object whose methods call into the native library described by `definitions`.
+///
+/// See the module-level documentation for the full registration pipeline.
+pub fn register(env: Env, handle: &LibraryHandle, definitions: JsObject) -> Result<JsObject> {
+    // Resolve rustbuffer symbols from definitions
+    let rb_ops = resolve_rustbuffer_symbols(handle, &definitions)?;
```
jhugman, line 130: Definitions has the names of the rust buffer symbols, the functions and callbacks vtables that are going to be called in to Rust. These will be generated in the `wrapper-ffi-player.ts` template.

runtimes/napi/src/register.rs
```diff
@@ -0,0 +1,610 @@
...
+    // Parse callback definitions if present
+    let callback_defs = parse_callbacks(&definitions)?;
+    let callback_defs = Rc::new(callback_defs);
+
+    // Parse struct definitions if present
+    let struct_defs = structs::parse_structs(&definitions)?;
+    let struct_defs = Rc::new(struct_defs);
+
+    let functions: JsObject = definitions.get_named_property("functions")?;
+    let mut result = env.create_object()?;
```
jhugman, line 141: Result is the object passed back in to Javascript which the API layer is going to call.

jhugman (commented):
Self-review.

runtimes/napi/src/ffi_c_types.rs
```diff
@@ -0,0 +1,96 @@
...
+/// Fields: `capacity` (allocated size), `len` (used size), `data` (heap pointer).
+/// The Rust side is responsible for allocation and deallocation; JS must call
+/// the library's `rustbuffer_free` to release buffers it receives.
+#[repr(C)]
+#[derive(Clone, Copy)]
+pub(crate) struct RustBufferC {
+    pub capacity: u64,
+    pub len: u64,
+    pub data: *mut u8,
+}
```
1egoman, line 38: What is the benefit to repeating this here over using the `RustBuffer` implementation in uniffi-rs [here](https://github.com/mozilla/uniffi-rs/blob/a2d6cee7ea0b427ad69dfbce536e25201ba1a9af/uniffi_core/src/ffi/rustbuffer.rs#L51-L62)?

Also I have a similar question for `ForeignBytes` but am assuming they probably were copied for the same reason.

runtimes/napi/src/fn_pointer.rs
```diff
@@ -0,0 +1,375 @@
...
+        // Call the C function pointer via libffi
+        let code_ptr = CodePtr::from_ptr(fn_ptr);
+        match &ret_type {
+            FfiTypeDesc::Void => {
+                unsafe { cif.call::<()>(code_ptr, &ffi_args) };
+                ctx.env.get_undefined().map(|v| v.into_unknown())
+            }
+            _ => Err(napi::Error::from_reason(format!(
+                "Non-void return types for fn_pointer_wrapper not yet supported: {ret_type:?}"
+            ))),
```
1egoman, line 370: Admittedly I haven't dug super deep into foreign calling since I never got that working in `uniffi-bindgen-node`, but how important of a case is this? [Reading through the relevant docs](https://github.com/mozilla/uniffi-rs/blob/main/docs/manual/src/internals/foreign_calls.md), I'm assuming that rust -> js functions calls would be using an out `RustBuffer` / `RustCallStatus` for return value handling - what are some use cases where a non `void` return would be needed?

runtimes/napi/src/napi_utils.rs
```diff
@@ -0,0 +1,220 @@
...
+///
+/// - `raw_env` must be a valid `napi_env` for the current callback scope.
+/// - `raw_val` must be a valid `napi_value` referring to a `TypedArray`.
+pub unsafe fn read_typedarray_data(
+    raw_env: napi::sys::napi_env,
+    raw_val: napi::sys::napi_value,
+) -> Option<(*const u8, usize)> {
+    let mut length: usize = 0;
+    let mut data: *mut c_void = std::ptr::null_mut();
+    let mut ab = std::ptr::null_mut();
```
1egoman, line 110: I think if this function were exposed with some minimal wrapping in the root `UniffiNativeModule` module as a utility, it might be a good first step to stripping out the `ref-napi` dependency being used in a few places at LiveKit to convert a node `Buffer` -> pointer and length.

I'm not sure if this suggestion is too premature but since it would be fairly trivial maybe it's worth considering?

runtimes/napi/src/register.rs
1egoman: I had mentioned [over here in the player proposal issue](https://github.com/jhugman/uniffi-bindgen-react-native/issues/360#issuecomment-4136978399) that it might be nice to be able to support conditional loading / unloading of dylibs (though IMO not critical for a proof od concept, but it was handy when dealing with a very large (~100mb) dylib). This is sort of "unrequiring" conceptually.

I bring this up because if it makes it easier to model everything as having a "load" and an "unload" step, that actually would IMO be the desired end state for node as well so treating it as a special case might be more trouble than its worth.

runtimes/napi/src/register.rs
```diff
@@ -0,0 +1,657 @@
...
+    let from_bytes_name: String = symbols
+        .get_named_property::<napi::JsString>("rustbufferFromBytes")?
+        .into_utf8()?
+        .as_str()?
+        .to_owned();
+
+    // We only need from_bytes and free at call time; alloc is resolved to validate it exists.
+    let _alloc_ptr = handle.lookup_symbol(&alloc_name)?;
+    let free_ptr = handle.lookup_symbol(&free_name)?;
+    let from_bytes_ptr = handle.lookup_symbol(&from_bytes_name)?;
```
1egoman, line 117: I'd be curious to hear your thinking here. For context, in `uniffi-bindgen-node`, I had difficulty getting `alloc` to work and ended up making the decision to create all `RustBuffers` via `from_bytes` just so I could get things to work.

I remember at the time not being a fan of this decision since it means your are effectively sending an extra copy of data across the ffi boundary for data that originates in js (rather than making the alloc call, and then filling the `RustBuffer` fully on the js end).

Is there some advantage I'm missing to using `from_bytes` exclusively?

1egoman (approved):
I generally follow what is going on here, great work!

I think next time we chat we should talk through how some of the rust -> js foreign calling logic works, I never had to dig into that for uniffi-bindgen-node so that's where most of the holes in my understanding of things are currently.

A suggestion a LLM made, which I thought might be worth considering:

> No `#[deny(unsafe_op_in_unsafe_fn)]`
> For a crate with this much unsafe, enabling `#[deny(unsafe_op_in_unsafe_fn)]` at the crate level would force each unsafe operation inside an unsafe fn to still be explicitly wrapped in an unsafe {} block, improving auditability. This is a relatively low-effort hardening step and aligns with the stated goal of minimizing the scope of each unsafe block.

runtimes/napi/src/ffi_c_types.rs
jhugman: The player itself (and none of the foreign bindings) have a dependency on or link to `uniffi_core`, so we need some `repr(C)` structs which match the ABI, but not be the dependency.

runtimes/napi/src/fn_pointer.rs
jhugman: Rust call status is an out param for calls in both directions. It contains an error buffer, which is a `RustBuffer`.
`RustCallStatus { err: RustBuffer, status: uint8 }`

For callbacks, and async callbacks, the return (and call status / error) is also an out param, rather than a straight forward return. So anything returns go through that path, not this one.

runtimes/napi/src/napi_utils.rs
jhugman: That may well be the direction we end up taking, however, we need to look after the `uniffi-rs` side as well.

https://github.com/mozilla/uniffi-rs/issues/2864 is the next step here.

jhugman:
@1egoman I like your idea of unloading / unrequiring. I have an [optimization/layering/hardening phase](379) to do, that would be a perfect fit for this.

runtimes/napi/src/register.rs
jhugman: I think `from_bytes` is the API provided for by uniffi to write into the Rust heap. I think you can probably do it with alloc and copying, but you'd have to do that from the foreign side; which becomes just another foot gun for every language's bindgen.

Either way, I think the number of copies that are made of the data remains the same.

runtimes/napi/src/fn_pointer.rs
1egoman: So just to confirm - it sounds like you are saying that there aren't any cases where the return value of a callback matters?

PR #371, opened 2026-03-31
Declare the package as CommonJS

### LINKED ISSUE #370: Exported values cannot be imported in Node
When using [the published NPM package of this repo](https://www.npmjs.com/package/uniffi-bindgen-react-native) as a Node dependency, attempting to import anything from it fails.

It's a problem whether the importing script uses CommonJS:
```javascript
#!/usr/bin/env node
// index.cjs
const r = require("uniffi-bindgen-react-native");
```

or ES modules:
```javascript
#!/usr/bin/env node
// index.mjs
import * as r from "uniffi-bindgen-react-native";
```

Either way, imports fail with this error:
```
Object.defineProperty(exports, "__esModule", { value: true });
                      ^

ReferenceError: exports is not defined in ES module scope
```

This is the same issue as #324, though in my experience it has not been fixed by #343.

AndrewFerr:
This allows exported values to be imported in Node.

Fixes #370

---

This is a simpler alternative to #325 which avoids having to add file extensions on imports.  It also works without having to re-emit JS with `tsc`.

The reason this works is because, as was noted in #324, the `tsconfig.json` is configured to emit CommonJS, and the `package.json`'s `type` field may simply be updated to match that.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "name": "James Hugman",
     "email": "james@hugman.tv"
   },
-  "type": "module",
+  "type": "commonjs",
   "main": "./typescript/dist/index.js",
   "types": "./typescript/dist/index.d.ts",
   "bin": {


jhugman:
Thanks so much for your contribution. Is the test failure related to the fix?

AndrewFerr:
> Is the test failure related to the fix?

So it is.  The failure is due to the generated WASM binding JS using ESM-style imports, which doesn't work with `"type": "commonjs"` modules:
```javascript
// examples/arithmetic/generated/wasm/ts/wasm-bindgen/index.js
import * as wasm from "./index_bg.wasm";
export * from "./index_bg.js";
import { __wbg_set_wasm } from "./index_bg.js";
__wbg_set_wasm(wasm);
wasm.__wbindgen_start();
```

AIUI, it's correct for the WASM bindings to use ESM, as they target web browsers, which do not support CommonJS.  

So it looks like setting the package's module format to CommonJS is the wrong approach for adding Node support, since web must remain a targeted platform.  That would make #325 the correct path forward, though I still don't know why the PR author was able to proceed without it once #343 was merged.

PR #375, opened 2026-04-06
Have typescript generate ESM code

alastaircoote:
Possible that I'm missing something here but when I try to use the module I receive an error that it contains CommonJS files inside a module that's `type: "module"`. And indeed that seems to be the case. I've tweaked the tsconfig to output ESM and everything is working fine for me locally.


jhugman:
Is this related to #370  ?

PR #377, opened 2026-04-07
Enable test fixtures to run a new NAPI flavour

jhugman:
This PR builds on #369 , to use the existing test fixtures to run on Node JS via a NAPI runtime player.

We add:

- a test run, in the `uniffi-fixture-testing` crate, using `tsx`.
- a new `wrapper-ffi-player.ts` pipeline to generate the `register` calls and types to be compatible with the existing `wrapper.ts`. This currently uses a very simple environment variable to resolve the binary, and only works on unixes.

We change:

- the `wrapper.ts` pipeline to thread the flavour through the builders, so we can generate (or not) the `ubrn_` prefix needed for WASM. We can start to remove this prefix from JSI now.
- each fixture with Napi variant to enable testing.

No attempt has been made to run anything outside of the fixtures. There is some code clean up needed in the generated code. This PR is about seeing the whole thing run.

No effort has been made to make it robust to which .so file is used, or to run on anything except macOS. All that is for another day.

To see this in action: 

```sh
(cd runtimes/napi/ && npm run build)
cargo test napi:: 2>&1 | grep -E "^(test |FAILED)" 
```

Then find a test fixture or example e.g. `examples/rondpoint-procmacro` and look in the:
- `tests/bindings/` directory to see the typescript test file.
- `generated/napi/ts` directory to see two generated files
- `src/lib.rs` file to see the Rust that is exercised.


crates/ubrn_bindgen/src/bindings/gen_typescript/api_module/builders.rs
```diff
@@ -47,11 +47,20 @@ pub(super) fn build_map(map: &general::MapType) -> TsSimpleWrapper {
...
+        format!("ubrn_{}", raw_name)
+    } else {
+        raw_name.to_string()
+    }
+}
+
 /// FFI function names match the `ffi_module` synthetic declarations.
 pub(super) fn build_string_helper(flavor: &AbiFlavor) -> TsStringHelper {
     // JSI has no global TextEncoder/TextDecoder, so it uses Rust-provided
     // string conversion functions instead.
```
1egoman: nitpick: The comment no longer matches with what the function below does - maybe something like the below could be better?

```suggestion
// Some flavors don't have a global TextEncoder/TextDecoder, so use
// Rust-provided string conversion functions instead.
```

crates/ubrn_bindgen/src/bindings/gen_typescript/templates/wrapper-ffi-player.ts
```diff
@@ -0,0 +1,108 @@
...
+    const libPath = process.env.UNIFFI_LIB_PATH;
+    if (!libPath) {
+      throw new Error(
+        "No library path available. " +
+        "Set the UNIFFI_LIB_PATH environment variable to the path of the shared library."
+      );
+    }
+    {%- endmatch %}
+    const mod_ = UniffiNativeModule.open(libPath);
+    _nativeModule = mod_.register(DEFINITIONS) as unknown as NativeModuleInterface;
```
1egoman, line 87: Out of curiosity, what is the return type of `mod_.register` prior to the cast? I scanned through https://github.com/jhugman/uniffi-bindgen-react-native/pull/377 again and it wasn't obvious where that was being defined. Is it just `any` because it's being imported from a js file (`uniffi-runtime-napi/lib.js`)?

1egoman (approved):
High level makes sense to me! I just noticed these small things scanning through but will give the tests a shot locally and potentially leave more comments if I run into unexpected behavior.

crates/ubrn_bindgen/src/bindings/gen_typescript/templates/wrapper-ffi-player.ts
jhugman: The [Rust `register` method](https://github.com/jhugman/uniffi-bindgen-react-native/pull/369/changes#diff-918c84833dbe86b6a9d0964c02cbe64ec6c2b391ccfa8c0e39291dae8431ceb7R128-R204) returns a `JsObject`, which I believe comes out as `any`.

crates/ubrn_bindgen/src/bindings/gen_typescript/api_module/builders.rs
jhugman: Hmm. The only backend that doesn't support text encoder is JSI, and that uses __C++__ string conversion functions. I'll change the comment.

PR #378, opened 2026-04-08
Optimizing strings and bytes over the FFI

jhugman:
The benchmarks I used were recording getting strings, arrays of strings and byte arrays to and from Rust.

I used string arrays as a proxy for strings that are part of larger types (from Option<String> to Records or Enums containing strings).

The headline numbers are quite surprising. On the same machine, sending strings across the FFI is much slower for JSI, but sending bytes is quicker for JSI.

As with all benchmarks, these numbers may not reflect real world usage.

The biggest speed ups were for JSI strings and string arrays; by reducing the number of JS/C++ calls per string per crossing:
- replacing decoding and encoding in directly into the a buffer from a string
- replacing the string encoding path with the native `TextEncoder`. **This requires react-native 0.74**.

### More detail

Functions are all shaped, e.g.:

```rust
// gets a string from Rust.
fn get_string(length: u32) -> String
// Rust takes a string and consumes it.
fn take_string(string: String) -> ()
```

## JSI

| Operation       | Before   | After    | Change |
|-----------------|----------|----------|--------|
| getString       | 487ms    | 406ms    | -17%   |
| takeString      | 666ms    | 623ms    | -6%    |
| getStringArray  | 3,287ms  | 2,192ms  | -33%   |
| takeStringArray | 6,900ms  | 3,092ms  | -55%   |
| getBytes        | 1,962ms  | ~1,913ms | noise  |
| takeBytes       | 4,292ms  | ~4,182ms | noise  |                                                                              
                                                          
  ## WASM                                                                                                                                     
                                                          
  | Operation       | Before  | After   | Change |
  |-----------------|---------|---------|--------|
  | getString       | 44ms    | 39ms    | -11%   |                                                                                            
  | takeString      | 384ms   | 397ms   | noise  |
  | getStringArray  | 35ms    | 25ms    | -29%   |                                                                                            
  | takeStringArray | 48ms    | 47ms    | noise  |                                                                                            
  | getBytes        | 5,698ms | 6,027ms | noise  |
  | takeBytes       | 6,162ms | 6,411ms | noise  |

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,6 +70,16 @@ jobs:
     name: 🧩 Integration tests (WASM bindings)
     runs-on: ubuntu-latest
     steps:
+      - name: Free up disk space
+        uses: jlumbroso/free-disk-space@main
+        with:
+          tool-cache: false
+          android: true
+          dotnet: true
+          haskell: true
+          large-packages: false
+          swap-storage: false
+
       - uses: actions/checkout@v4
 
       - name: Installing
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1972,6 +1972,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "uniffi-fixture-benchmark"
+version = "0.1.0"
+dependencies = [
+ "ubrn_fixture_testing",
+ "ubrn_macros",
+ "uniffi",
+]
+
+[[package]]
 name = "uniffi-fixture-callbacks"
 version = "0.22.0"
 dependencies = [
--- a/cpp/includes/UniffiString.h
+++ b/cpp/includes/UniffiString.h
@@ -12,9 +12,23 @@ namespace uniffi_jsi {
 using namespace facebook;
 
 template <> struct Bridging<std::string> {
-  static jsi::Value arraybuffer_to_string(jsi::Runtime &rt,
-                                          const jsi::Value &value) {
+  static jsi::Value string_from_buffer(jsi::Runtime &rt,
+                                       const jsi::Value &value) {
     try {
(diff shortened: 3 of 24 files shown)


zzorba (commented):
so I'm not sure why, but when I try to run against this commit locally (after rebuilding), all three platforms (web/android/ios) complain about `uniffiTypeNameSymbol` being undefined.

I don't see anything in this change that might explain that though, previously we were running 0.31.0-2.

This change looks good, though I would love to run it in our app before giving the 👍

jhugman:
> This change looks good, though I would love to run it in our app before giving the 👍

Sorry, it took me a while to get there.  I've run the app now, without any issues.

zzorba (dismissed):
LGTM, change is working great with that enum addition

zzorba (commented):
actually, I'm seeing a change in type signatures with this branch.

This enum:

```
#[derive(Clone, uniffi::Enum)]
pub enum EditedContent {
    RoomMessage {
        content: Arc<RoomMessageEventContentWithoutRelation>,
    },
    MediaCaption {
        caption: Option<String>,
        formatted_caption: Option<FormattedBody>,
        mentions: Option<Mentions>,
    },
    PollStart {
        poll_data: PollData,
    },
},
```

Is generating with Mention as required, not | undefined

```

    constructor(inner: {
      caption: string;
      formattedCaption: FormattedBody;
      mentions: Mentions;
    }) {
      super("EditedContent", "MediaCaption");

      this.inner = Object.freeze(inner);
    }
```

Looks like the Option is not being respected here?

zzorba (approved):
Thanks for the enum fix James, branch looks good and is working great!

PR #380, opened 2026-04-10
Commit auto-applied change to the lockfile

AndrewFerr:
This change gets applied to Cargo.lock when building, so check it into version control.


jhugman (approved):

jhugman:
@AndrewFerr thank you so much for doing this. It's been annoying me too.

PR #381, opened 2026-04-16
Remove redundant files from the package bundle

Simek:
# Why

When working on bundle code explorer for React Native Directory I have spotted that there are several files in the bundle, which can be safely skipped.

# How

Update `.npmignore` to skip bundling the redundant files: `.github` directory content and development tools configs.


jhugman (approved):
Nice!

Thank you so much for your contribution!

Issue #379, opened 2026-04-09
Layer refactor of NAPI Player


jhugman:
There are three objectives to this:

- Separating the NAPI player in to a "frontend" and "backend". The frontend talks NAPI, the backend talks uniffi. This will put the player on a good footing for optimization and re-using for other players.
- Optimize the backend: a lot of the transforming of arguments, returns etc per type that is currently being done on a per call basis can be done once at registration time.
- Better [lifecycle handling][ryan], for unloading dylibs, and hot-reloading. 

[ryan]: https://github.com/jhugman/uniffi-bindgen-react-native/pull/369#discussion_r3054208664_

jhugman:
Closed by #385.

PR #382, opened 2026-04-16
Enable all fixtures for Napi

jhugman:
This PR fixes the `runtimes/napi` player to run the remaining fixtures.

The struct helper functions (e.g. `uniffi_free` and `uniffi_clone` being called from C++ and Rust were in camelCase, they've now switched to snake case. This meant changing C++ and Rust in two places, and typescript.

The second major fix was lining up the callbacks and async callbacks.

Finally, there was some work to get it working on Linux. This initiated the beginnings of [the lifecylce/layer refactor](#379). 

---

#### callbacks

The vtable trampoline (`structs.rs`) had a single code path for handling the JS callback's return value. The problem was that uniffi uses two different calling conventions depending on the function:

- **Pass-by-reference status**: the RustCallStatus is passed as a mutable arg, JS mutates it in place, Rust reads it back.
- **UniffiResult / out-pointer protocol**: the callback returns `{ code, pointee?, errorBuf? }` as a JS object; Rust reads the return value and status from that object.

These two paths were tangled together. The fix split them into explicit branches: if `out_return` is true, extract code/pointee/errorBuf from the returned JS object; otherwise, read back the mutated status arg. This got the callbacks fixture passing.

#### async-callbacks, futures, arc-futures

Two more gaps exposed by the `async-callbacks` fixtures:

1. **`RustCallStatus` as an inline struct field** — async callback trampolines return a struct that *contains* a `RustCallStatus`. The previous code assumed status was always a separate pointer arg. Added handling for this case in the `out_return` branch: when the return type has `has_rust_call_status` but it's embedded in the struct, unpack it from the JS result object alongside the pointee.

1. **Callback-typed struct fields in return values** — some structs returned from callbacks contain vtable function pointers (i.e., callbacks themselves). The new `marshal_js_struct_to_c_bytes` function handles this: it walks the struct definition field by field, marshalling `UInt64`/`Handle` fields as BigInt and `Callback` fields by minting new libffi closures on the spot.

### DIFF
--- a/crates/ubrn_bindgen/src/bindings/gen_typescript/api_module/builders.rs
+++ b/crates/ubrn_bindgen/src/bindings/gen_typescript/api_module/builders.rs
@@ -631,7 +631,7 @@ fn build_vtable_field(
     ffi_fn_types: &HashMap<String, &general::FfiFunctionType>,
     flavor: &AbiFlavor,
 ) -> TsVtableField {
-    let name = fn_name(&vm.callable.name);
+    let name = vm.callable.name.clone();
     let method = Some(build_callable(config, &vm.callable, &None, None, flavor));
 
     let general::FfiType::Function(ref ffi_fn_name) = vm.ffi_type.ty else {
--- a/runtimes/napi/src/structs.rs
+++ b/runtimes/napi/src/structs.rs
@@ -226,6 +226,10 @@ pub unsafe extern "C" fn vtable_trampoline_callback(
     args: *const *const c_void,
     userdata: &VTableTrampolineUserdata,
 ) {
+    if crate::is_shutting_down() {
+        return;
+    }
+
     // Zero-initialize the result buffer so that early returns (from error paths
     // in the main-thread or cross-thread trampolines) produce a deterministic
     // zero value rather than leaving uninitialized memory that the Rust caller
@@ -1389,6 +1393,16 @@ pub fn create_callback_closure(
         )?
     };
     let mut tsfn = tsfn;
+
+    // Register the raw TSFN handle so the env cleanup hook can abort it at shutdown.
+    // This ensures that any Rust worker threads blocked on `rx.recv()` are unblocked
+    // when Node.js exits, preventing SIGSEGV on Linux from `dlclose()` racing with
+    // threads that still hold function pointers into the unloaded library.
+    crate::register_tsfn(tsfn.raw());
+
+    // Unref the TSFN so it does not prevent the Node.js event loop from exiting.
+    // The env cleanup hook will abort it if the process shuts down while Rust
+    // worker threads are still waiting.
     tsfn.unref(env)?;
 
     // SAFETY: `userdata_ptr` is a valid, uniquely-owned heap allocation.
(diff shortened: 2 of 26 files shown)


crates/ubrn_bindgen/src/bindings/gen_typescript/api_module/builders.rs
```diff
@@ -564,7 +564,7 @@ fn build_vtable_field(
     ffi_fn_types: &HashMap<String, &general::FfiFunctionType>,
     flavor: &AbiFlavor,
 ) -> TsVtableField {
-    let name = fn_name(&vm.callable.name);
+    let name = vm.callable.name.clone();
```
jhugman, line 634: This one line accounts for the first part of this PR: moving the struct callbacks and `unffiFree` and `uniffiClone` to snake case.

Now the model is: 

rust calls `callback_method` in the generated API `wrapper.ts` file, and `callback_method` then calls the clients actual object method `callbackMethod`.

jhugman (commented):
Self review. The getting-to-work-with-linux is likely going to be a theme of difficulty for this project. 

I am going to attempt #379 after this one to make that a bit more maintainable.

typescript/src/async-rust-call.ts
```diff
@@ -80,6 +80,19 @@ export async function uniffiRustCallAsync<F, S extends UniffiRustCallStatus, T>(
...
+  // don't prevent process exit). Without a ref'd handle here, the event loop may
+  // exit before the callback fires — especially on Linux where event loop
+  // draining is more aggressive than on macOS.
+  // Only needed when setInterval is natively available (i.e. Node.js / napi
+  // runtime); Hermes/JSI provides timers via a polyfill that may not be loaded
+  // yet at import time, and WASM has its own event loop.
+  const keepAlive =
+    typeof setTimeout === "function"
+      ? setTimeout(() => {}, 2_147_483_647)
+      : undefined;
```
1egoman: nitpick: The comment references setInterval but the implementation uses setTimeout. Should the comment be updated to match the implementation?

runtimes/napi/src/structs.rs
```diff
@@ -1184,3 +1311,112 @@ pub fn build_vtable_struct(
...
+        raw_env: env.raw(),
+        fn_ref,
+        arg_types: cb_def.args.clone(),
+        ret_type: cb_def.ret.clone(),
+        has_rust_call_status: cb_def.has_rust_call_status,
+        out_return: cb_def.out_return,
+        tsfn: None,
+        rb_ops: *rb_ops,
+        callback_defs: callback_defs.clone(),
+        struct_defs: struct_defs.clone(),
```
1egoman, line 1377: nitpick: Would it be better to share these `HashMap`s here by putting them in an `Arc` or similar? I'd imagine these could get fairly large.

1egoman (approved):

typescript/src/async-rust-call.ts
jhugman: Yes, you're right!

I've edited the comment.

runtimes/napi/src/structs.rs
jhugman: Yes: however: this will all be optimized away soon.

I've switched to a combination of `&HashMap<_, _>` and `Arc<HashMap<_, _>>` in the meantime.

PR #383, opened 2026-04-17
feat: add global configuration which enables emitting `Vec<u8>` as `Uint8Array` in TS

### LINKED ISSUE #326: `Vec<u8>` is compiled to `ArrayBuffer`, where it should be compiled to `Uint8Array`
Semantically these two types are distinct: [`ArrayBuffer`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer) is just a memory region and can be the underlying storage for just about anything really, whereas [`Uint8Array`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array) is specifically a byte vector. 

As such, `Vec<u8>` should transform to the narrower type, and not make downstream consumers of the compiled Rust code perform those conversions manually.

coriolinus:
For some teams (such as my own) this will be an improvement over rendering them as `ArrayBuffer`.

- Closes #326.

Technique as recommended [here](https://github.com/jhugman/uniffi-bindgen-react-native/issues/326#issuecomment-4127259938).

### DIFF
--- a/crates/ubrn_bindgen/src/bindings/gen_typescript/api_module/builders.rs
+++ b/crates/ubrn_bindgen/src/bindings/gen_typescript/api_module/builders.rs
@@ -11,35 +11,36 @@ use std::collections::HashMap;
 use heck::{ToLowerCamelCase, ToUpperCamelCase};
 use uniffi_bindgen::pipeline::general;
 
+use crate::bindings::gen_typescript::Config;
 use crate::{bindings::gen_typescript::config::CustomTypeConfig, switches::AbiFlavor};
 
 use super::docstring::{format_docstring, format_docstring_indented};
 use super::nodes::*;
 use super::type_helpers::*;
 
-pub(super) fn build_optional(opt: &general::OptionalType) -> TsSimpleWrapper {
+pub(super) fn build_optional(cfg: &Config, opt: &general::OptionalType) -> TsSimpleWrapper {
     TsSimpleWrapper {
         infra_class: "FfiConverterOptional".into(),
         ffi_converter_name: ffi_converter_name_for(&opt.self_type),
-        type_label: type_label_for(&opt.self_type.ty),
+        type_label: type_label_for(cfg, &opt.self_type.ty),
         inner_converters: vec![ffi_converter_name_for(&opt.inner)],
     }
 }
 
-pub(super) fn build_sequence(seq: &general::SequenceType) -> TsSimpleWrapper {
+pub(super) fn build_sequence(cfg: &Config, seq: &general::SequenceType) -> TsSimpleWrapper {
     TsSimpleWrapper {
         infra_class: "FfiConverterArray".into(),
         ffi_converter_name: ffi_converter_name_for(&seq.self_type),
-        type_label: type_label_for(&seq.self_type.ty),
+        type_label: type_label_for(cfg, &seq.self_type.ty),
         inner_converters: vec![ffi_converter_name_for(&seq.inner)],
     }
 }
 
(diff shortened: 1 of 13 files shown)


crates/ubrn_bindgen/src/bindings/gen_typescript/api_module/builders.rs
```diff
@@ -11,38 +11,39 @@ use std::collections::HashMap;
...
 
+use crate::bindings::gen_typescript::Config;
 use crate::{bindings::gen_typescript::config::CustomTypeConfig, switches::AbiFlavor};
 
 use super::docstring::{format_docstring, format_docstring_indented};
 use super::nodes::*;
 use super::type_helpers::*;
 
-pub(super) fn build_optional(opt: &general::OptionalType) -> TsSimpleWrapper {
+pub(super) fn build_optional(cfg: &Config, opt: &general::OptionalType) -> TsSimpleWrapper {
```
jhugman: Nit: The Great Vowel Shortage of 1986 left scars on us all, friend. It's safe to call it `config` now I think. 😃

crates/ubrn_bindgen/src/bindings/gen_typescript/api_module/builders.rs
```diff
@@ -72,16 +73,17 @@ pub(super) fn build_string_helper(flavor: &AbiFlavor) -> TsStringHelper {
 
 /// `config` = `None` means no custom lift/lower; falls back to the builtin converter.
 pub(super) fn build_custom_type(
+    cfg: &Config,
     custom: &general::CustomType,
-    config: Option<&CustomTypeConfig>,
+    type_config: Option<&CustomTypeConfig>,
```
jhugman: Does this disappear now? isn't `type_config` a member of `Config`?

crates/ubrn_bindgen/src/bindings/gen_typescript/api_module/type_helpers.rs
```diff
@@ -59,7 +66,13 @@ pub(super) fn type_label_for(ty: &general::Type) -> String {
...
         general::Type::Boolean => "boolean".into(),
         general::Type::String => "string".into(),
-        general::Type::Bytes => "ArrayBuffer".into(),
+        general::Type::Bytes => {
+            if cfg.strict_byte_arrays {
+                "Uint8Array".into()
+            } else {
+                "ArrayBuffer".into()
+            }
+        }
```
jhugman, line 75: This is what all the threading was for.

jhugman (approved):
Nice! Thank you for perservering with this, and giving such a complete, pain free PR.

I have one naming nit, but not nothing traumatic.

crates/ubrn_bindgen/src/bindings/gen_typescript/api_module/builders.rs
coriolinus: Fair enough; 32b5dbb

crates/ubrn_bindgen/src/bindings/gen_typescript/api_module/builders.rs
coriolinus: Oh, good spot, I missed that. https://github.com/jhugman/uniffi-bindgen-react-native/pull/383/commits/aa1cc57ab7035e11dfad47435de276fc0991964b

coriolinus:
Thanks for the quick approval! I've addressed both your comments.

I appear not to have merge permissions on this repo--I don't have a merge button, even greyed out--so I'll depend on you to merge this when you're satisfied that this is ready.

PR #386, opened 2026-04-23
Add ci to publish uniffi-runtime-napi to node.

jhugman:


runtimes/napi/package.json
```diff
@@ -1,21 +1,36 @@
...
+    "README.md"
+  ],
   "scripts": {
-    "build": "napi build --platform --release",
-    "build:debug": "napi build --platform",
+    "build": "napi build --platform --release --js-package-name @uniffi-runtime/napi",
+    "build:debug": "napi build --platform --js-package-name @uniffi-runtime/napi",
+    "artifacts": "napi artifacts --js-package-name @uniffi-runtime/napi",
+    "create-npm-dirs": "napi create-npm-dirs",
+    "prepublishOnly": "napi prepublish -t npm",
```
SimonThormeyer: Just a drive-by comment: have you considered adding a `prepare` script (which could just do `npm run build`), to allow this player being installed as a git dependency?

Not entirely sure of the consequences of this though.

.github/workflows/napi-publish.yml
```diff
@@ -0,0 +1,166 @@
...
+            runner: windows-latest
+
+    name: Build ${{ matrix.target }}
+    runs-on: ${{ matrix.runner }}
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 20
```
1egoman: nitpick: You maybe should consider using node v24, github actions is in the midst of deprecating node v20.

(There's a few more instances in the file, I'll just leave this comment once.)

1egoman (approved):

runtimes/napi/package.json
jhugman: Interesting, no I hadn't considered it.

What changes would you suggest? (I must admit, I'm not really a node developer).

SimonThormeyer: I would suggest adding 

```
"prepare": "npm run build",
```

Note that this suggestion is merely based on the need of our project to occasionally track ubrn's main directly and these docs
https://docs.npmjs.com/cli/v8/using-npm/scripts#life-cycle-scripts, I don't have any experience with actually using it.

The docs say
> NOTE: If a package being installed through git contains a prepare script, its dependencies and devDependencies will be installed, and the prepare script will be run, before the package is packaged and installed.

jhugman: Ok; I went down a little rabbit hole. We're now generating a root package by copying to `npm/root` and publishing from there, so your prepare script _should_ work.

.github/workflows/napi-publish.yml
jhugman: Updated the node version, and the actions to their latest versions (as per https://simonw.github.io/actions-latest/versions.txt)

PR #389, opened 2026-04-29
Fix/reserved C++ words template checking

### LINKED ISSUE #387: Handle TypeScript reserved keywords in generated bindings
Hi, while updating the Ferrostar React Native implementation, I ran into an issue with TypeScript binding generation when a Rust method name matches a TypeScript reserved keyword.

For example:
```rust
pub trait NavigationCache: Send + Sync {
    fn save(&self, record: Vec<u8>);
    fn load(&self) -> Option<Vec<u8>>;
    fn delete(&self);
}
```

Since delete is a reserved keyword in TypeScript, generating bindings for this trait fails. 

It would be nice if the binding generator could detect reserved keywords and automatically rename them to a safe alternative in the generated TypeScript, instead of producing invalid code.

I don't have the exact error the generator output at the moment. But I can grab it if need be.

bjtrounson:
This PR should fix #387. We found the issue wasn't related to the typescript side since it does check the reserved words. But actually the C++ side where it wasn't doing the same check. I have added some similar functions to the C++ template generation the should emulate the same behavior the typescript generator does. 

I'm not that great at the C++ side of things, so if there is something wrong with my implementation let me know. 

I have added a test to the test_callbacks_regressions.ts. I don't know if this is the best place but I thought since the original example was from the callback area, and it's one test it would be a bit pointless to separate it.

### DIFF
--- a/crates/ubrn_bindgen/src/bindings/gen_cpp/filters.rs
+++ b/crates/ubrn_bindgen/src/bindings/gen_cpp/filters.rs
@@ -3,10 +3,110 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/
  */
+use std::collections::HashSet;
+use std::sync::LazyLock;
+
 use heck::{ToLowerCamelCase, ToUpperCamelCase};
 use uniffi_bindgen::{interface::FfiType, ComponentInterface};
 
 use super::extensions::{CppComponentInterfaceExt as _, CppFfiTypeExt as _};
+
+static CPP_KEYWORDS: LazyLock<HashSet<&'static str>> = LazyLock::new(|| {
+    HashSet::from([
+        "alignas",
+        "alignof",
+        "and",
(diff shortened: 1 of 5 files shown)


crates/ubrn_bindgen/src/bindings/gen_cpp/filters.rs
```diff
@@ -74,8 +174,28 @@ pub fn ffi_type_name(ffi_type: &FfiType) -> Result<String, askama::Error> {
...
+    Ok(rewrite_cpp_keywords(nm.to_string()))
+}
+
+pub fn cpp_arg_name(nm: &str) -> Result<String, askama::Error> {
+    cpp_ident(nm)
+}
+
+pub fn cpp_field_name(nm: &str) -> Result<String, askama::Error> {
+    cpp_ident(nm)
+}
```
jhugman, line 195: Minorest of minor nits: the `cpp_` prefix seems redundant here.

jhugman (approved):
This is great! Thank you so much!

PR #390, opened 2026-05-05
NAPI: Resolve the location of the cdylib file, controlled at bindgen time

jhugman:
This PR introduces a resolve-lib module into the NAPI runtime.

This looks after the finding of the correct filename for the platform (e.g. `.so` for linux, `.dll` for windows, etc) and finding it on disk.

Three strategies:
- absolute: use an absolute path to the cdylib file.
- colocated: located next to the generated file
- use npm: use require to look up the path to a file installed via NPM.

### DIFF
--- a/crates/ubrn_bindgen/src/bindings/gen_typescript/ffi_module_player/nodes.rs
+++ b/crates/ubrn_bindgen/src/bindings/gen_typescript/ffi_module_player/nodes.rs
@@ -4,6 +4,21 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/
  */
 
+use camino::Utf8PathBuf;
+
+/// How the generated player should locate the cdylib at load time.
+///
+/// Maps 1:1 to the three resolveLibPath modes in `@uniffi-runtime/napi`.
+#[derive(Clone, Debug)]
+pub enum LibResolution {
+    /// Look for the conventional filename next to the binding.
+    Colocated,
+    /// Bake an absolute path into the generated code.
+    Absolute(Utf8PathBuf),
+    /// Resolve via `<base>-<triple>` platform packages.
+    Require(String),
+}
+
 /// IR for the player-style `{namespace}-ffi.ts`.
 ///
 /// Generates a `DEFINITIONS` object for the napi player's `register()` call,
@@ -11,9 +26,10 @@
 pub(crate) struct PlayerFfiModule {
     /// Whether to suppress `@ts-nocheck` for strict type checking.
     pub strict_type_checking: bool,
-    /// Optional library path baked into the generated code.
-    /// If None, the getter accepts a path argument at runtime.
-    pub lib_path: Option<String>,
+    /// The crate name, passed to `resolveLibPath` so error messages name it.
+    pub crate_name: String,
+    /// How the player should locate the library at runtime.
+    pub lib_resolution: LibResolution,
     /// Rustbuffer management symbol names.
     pub symbols: PlayerSymbols,
     /// FFI function registrations for `register({ functions: { ... } })`.
--- a/crates/ubrn_bindgen/src/bindings/gen_typescript/templates/wrapper-ffi-player.ts
+++ b/crates/ubrn_bindgen/src/bindings/gen_typescript/templates/wrapper-ffi-player.ts
@@ -7,8 +7,8 @@
 // @ts-nocheck
 {%- endif %}
 
-import lib from "@uniffi-runtime/napi/lib.js";
-const { UniffiNativeModule, FfiType } = lib;
+import lib from "@uniffi-runtime/napi";
+const { UniffiNativeModule, FfiType, resolveLibPath } = lib;
 
 import {
   type StructuralEquality as UniffiStructuralEquality,
@@ -71,18 +71,17 @@ interface NativeModuleInterface {
 let _nativeModule: NativeModuleInterface | undefined;
 const getter: () => NativeModuleInterface = () => {
   if (!_nativeModule) {
-    {%- match module.lib_path %}
-    {%- when Some with (path) %}
-    const libPath = "{{ path }}";
-    {%- when None %}
-    const libPath = process.env.UNIFFI_LIB_PATH;
-    if (!libPath) {
-      throw new Error(
-        "No library path available. " +
-        "Set the UNIFFI_LIB_PATH environment variable to the path of the shared library."
-      );
-    }
-    {%- endmatch %}
+    const libPath = resolveLibPath({
+      crateName: "{{ module.crate_name }}",
+      callerUrl: import.meta.url,
+      {%- match module.lib_resolution %}
+      {%- when LibResolution::Colocated %}
+      {%- when LibResolution::Absolute with (path) %}
+      override: "{{ path }}",
+      {%- when LibResolution::Require with (base) %}
+      npmPackageBase: "{{ base }}",
+      {%- endmatch %}
+    });
     const mod_ = UniffiNativeModule.open(libPath);
     _nativeModule = mod_.register(DEFINITIONS) as unknown as NativeModuleInterface;
   }
--- a/runtimes/napi/typescript/src/resolve-lib.ts
+++ b/runtimes/napi/typescript/src/resolve-lib.ts
@@ -1,0 +1,187 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/
+ */
+
+import { existsSync } from "node:fs";
+import { createRequire } from "node:module";
+import { dirname, isAbsolute, join } from "node:path";
+import { fileURLToPath } from "node:url";
+
+interface ProcessLike {
+  platform: NodeJS.Platform;
+  arch: string;
+  report?: { getReport(): { header?: { glibcVersionRuntime?: string } } };
+}
+
+function detectTriple(proc: ProcessLike): string {
+  const { platform, arch } = proc;
+  if (platform === "darwin" && arch === "arm64") return "darwin-arm64";
+  if (platform === "darwin" && arch === "x64") return "darwin-x64";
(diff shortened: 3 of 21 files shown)


crates/ubrn_bindgen/src/bindings/gen_typescript/templates/wrapper-ffi-player.ts
```diff
@@ -71,18 +71,17 @@ interface NativeModuleInterface {
...
+      crateName: "{{ module.crate_name }}",
+      callerUrl: import.meta.url,
+      {%- match module.lib_resolution %}
+      {%- when LibResolution::Colocated %}
+      {%- when LibResolution::Absolute with (path) %}
+      override: "{{ path }}",
+      {%- when LibResolution::Require with (base) %}
+      npmPackageBase: "{{ base }}",
+      {%- endmatch %}
+    });
```
jhugman, line 84: This is where library resolution is selected at runtime.

runtimes/napi/typescript/src/resolve-lib.ts
```diff
@@ -0,0 +1,187 @@
+/*
```
jhugman, line 1: This is the file that does the work of finding the correct triple and file location.

The mechanism was almost entirely lifted from uniffi-bindgen-node.

crates/ubrn_bindgen/src/bindings/gen_typescript/ffi_module_player/nodes.rs
```diff
@@ -4,16 +4,32 @@
...
     pub strict_type_checking: bool,
-    /// Optional library path baked into the generated code.
-    /// If None, the getter accepts a path argument at runtime.
-    pub lib_path: Option<String>,
+    /// The crate name, passed to `resolveLibPath` so error messages name it.
+    pub crate_name: String,
+    /// How the player should locate the library at runtime.
+    pub lib_resolution: LibResolution,
     /// Rustbuffer management symbol names.
     pub symbols: PlayerSymbols,
```
1egoman, line 34: I'm not sure I understand why `lib_resolution` is at this level - wouldn't this be napi player specific? And other future players (wasm, jsi, etc) would not need this?

Or is the thinking you're adding it here for now and will further abstract once other players exist in the future?

runtimes/napi/typescript/src/resolve-lib.ts
```diff
@@ -0,0 +1,187 @@
...
+      proc.report?.getReport()?.header?.glibcVersionRuntime !== undefined;
+    return `linux-${arch}-${isGnu ? "gnu" : "musl"}`;
+  }
+  if (platform === "win32" && (arch === "x64" || arch === "arm64")) {
+    return `win32-${arch}-msvc`;
+  }
+  throw new Error(
+    `Unsupported platform/arch combination: ${platform}/${arch}. ` +
+      `Supported: darwin-{arm64,x64}, linux-{x64,arm64}-{gnu,musl}, win32-{x64,arm64}-msvc.`,
+  );
```
1egoman, line 33: Is there value in requiring these exact platform configurations? I realize these are all the fairly common ones, but at least with my uniffi-bindgen-node implementation I had tried to give the user a lot more control what platforms they wanted to build and how each package mapped to which platforms.

runtimes/napi/typescript/src/resolve-lib.ts
```diff
@@ -0,0 +1,187 @@
...
+      message:
+        `Could not find platform package for crate "${crateName}": tried "${pkgName}". ` +
+        `Detected platform: ${triple}. Either the package is not installed (run npm install) ` +
+        `or this platform is not in the published matrix.`,
+      mode: "npmPackageBase",
+      crateName,
+      attempted: [pkgName],
+      cause,
+    });
+  }
```
1egoman, line 170: Similar to my above comment - I wonder if it makes sense for this to be more customizable.

A use case I had thought about in the uniffi-bindgen-node version of this logic: Lets say you wanted to do something very custom with one of these "dylib containing" packages which isn't configurable within the guardrails outlined here. With the way this was done in uniffi-bindgen-node, the `pkgName` is fully user configurable, so they could (for example) point at a custom package for one set of platform attributes (`os`, `arch`, etc) which could be a "bodge layer" and do their custom logic, which would then proxy to the actual underlying generated package.

The con to this is you end up with a large nested switch / if+elseif statement, handling all combinations of platform attributes individually. However, that seems to be fairly par for the course for these types of tools, for example, see `./native.js` in https://www.npmjs.com/package/@livekit/rtc-ffi-bindings?activeTab=code for an example of what this big switch looks like in a `napi-rs` generated package.

runtimes/napi/typescript/src/resolve-lib.ts
```diff
@@ -0,0 +1,187 @@
...
+  if (platform === "win32") return `${crateName}.dll`;
+  const ext = platform === "darwin" ? "dylib" : "so";
+  return `lib${crateName}.${ext}`;
+}
+
+export function resolveLibPath(opts: ResolveLibPathOptions): string {
+  if ("override" in opts && opts.override !== undefined) {
+    return resolveOverride(opts.crateName, opts.override);
+  }
+  if ("npmPackageBase" in opts && opts.npmPackageBase !== undefined) {
```
1egoman, line 91: You might consider adding some sort of environment variable override here, like `${opts.crateName.toUpperCase()}_PATH` or something, which would let users manually change the dylib resolution themselves for debugging / development / etc.

Related: https://github.com/livekit/uniffi-bindgen-node/pull/28

runtimes/napi/typescript/src/resolve-lib.ts
```diff
@@ -0,0 +1,187 @@
...
+      crateName,
+      attempted: [pkgName],
+      cause,
+    });
+  }
+
+  const binaryPath = join(
+    dirname(pkgJsonPath),
+    libFileName(crateName, process.platform),
+  );
```
1egoman, line 175: This approach of looking for a file directly in the package is fairly similar to what `napi-rs` does and I think works for in the majority of cases.

However, with uniffi-bindgen-node, I opted to instead make the generated package actually contain a small javascript shim which contained a default export of a function which when called would return a value of type `{ triple: string, path: string }` (the `triple` was not being used, that could be dropped, but I thought it could be handy in the future).

The reason I did this was (like some of my other earlier suggestions) to facilitate customization. For example, if a user wanted to do something like automatically download a dylib from a URL somewhere rather than include the binary in the package / maybe do some sort of preprocessing on the dylib file before loading / etc, then they could use this function as a "hook" to tap into the loading process and do whatever they want.

IMO, unless there's a really strong argument against exposing a small js shim in these dependent packages, I think it adds a lot of futureproofing opportunity and acts as a great extension point.

runtimes/napi/typescript/src/resolve-lib.ts
```diff
@@ -0,0 +1,187 @@
...
+  return prev;
+}
+
+function resolveNpmPackage(
+  crateName: string,
+  callerUrl: string,
+  npmPackageBase: string,
+): string {
+  const triple = _detectTripleImpl();
+  const pkgName = `${npmPackageBase}-${triple}`;
```
1egoman, line 153: It's possible I'm missing this piece - but how do these packages get generated right now?

For reference, in `uniffi-bindgen-node`, I added a (poorly named 😬 ) subcommand which could take a triple and a source and generate a standards compliant "dylib package":
```
$ uniffi-bindgen-node publishing-scaffold-native-package --help
Generates a template for a npm package that encapsulates a built dll / dylib / dll.                                                                

The module has a default export of a function, which when called returns an object with two keys:
- "triple" mapping to the built rust triple the package represents       
- "path" containing an absolute path to the dll / dylib / so that is included in the package.
Typescript definitions are also included. ie:
                                                                             
> require('./example-native-package').default()
{                         
  path: '/path/to/example-native-package/src/libplugins_ai_coustics_uniffi.dylib',
  triple: 'aarch64-apple-darwin'
}          

Only intended for use when publishing package to npm for distribution.
                                      
Usage: uniffi-bindgen-node publishing-scaffold-native-package [OPTIONS] --package-name <PACKAGE_NAME> <LIB_SOURCE> <LIB_TRIPLE>
                                      
Arguments:                                                                   
  <LIB_SOURCE>
          Path to the compiled library (.so, .dylib, or .dll)

  <LIB_TRIPLE>
          Rust triple representing the platform that `lib-source` was compiled under
                                      
Options:
      --lib-package-name <LIB_PACKAGE_NAME>
          The value of the "name" field in the main package's "package.json". Used to add context to the README.md about the purpose of the package

      --package-name <PACKAGE_NAME>
          The value to set the "name" field of the generated package.json

      --package-version <PACKAGE_VERSION>
          The value to set the "version" field of the generated package.json
          
          [default: 0.0.0]

      --package-os <PACKAGE_OS>
          

      --package-cpu <PACKAGE_CPU>
          

  -o, --out-dir <OUT_DIR>
          Output directory to write the native package into
          
          [default: ./scaffolded-native-package]

  -h, --help
          Print help (see a summary with '-h')
```

While I think it some parts of this could use some additional polish, just generally this subcommand approach has worked quite well in practice, especially in ci, since this separate command can be run at the end of each platform specific build pipeline + the package saved as an artifact, then at the very end at publish time all these artifacts can be rehydrated and published concurrently.

1egoman (commented):
Generally looks pretty good! Since this is something I've done quite a bit of thinking about, I've left some fairly in depth comments - curious your take on all of them.

One other thing - obviously not important right now, but I think long term it would be pretty sweet if there was a github action which could be used to automate some of the packaging complexity. I think this could be something that is developed at LiveKit though and eventually could make its way here once we have some proven patterns that we've settled on.

jhugman:
I think where I'm at right now: either we do the publishing story in this PR, so we can get the vertically integrated end-to-end story, OR we land this now, in its imperfect state, and then iterate it when we do the packaging strory.

1egoman:
@jhugman I think I'm fine with the latter - merge this now, and further iterate as part of this packaging work. Assuming that sounds good, I'll approve this.

1egoman (approved):
Approving to unblock merging assuming that some of the things I surfaced get thought about more in a follow up

PR #391, opened 2026-05-06
Add index.ts generation to player pipeline

jhugman:
This adds a version of the `index.ts` generation taken from `ubrn_cli` and uses it to generate `index.ts` for pipelines using a runtime (currently just NAPI). It unifies sync and async calls.

This is generated in the ts out-directory by the generate command.

It is tested in the `ext-types` fixture, by a second test script which imports the Rust from the generated index.ts.


crates/ubrn_bindgen/src/bindings/gen_typescript/templates/index.ts
```diff
@@ -0,0 +1,38 @@
...
+{% if flavor.supports_sync_initialization() %}
+let initialized = false;
+if (!initialized) {
+  {%- for m in modules %}
+  {{ m.ts() }}.default.initialize();
+  {%- endfor %}
+  initialized = true;
+}
+
+export async function uniffiInitAsync() { /* sync flavor: no-op */ }
```
1egoman, line 19: Should this log a warning if called?

1egoman (approved):

crates/ubrn_bindgen/src/bindings/gen_typescript/templates/index.ts
jhugman: No. I think the client code should be able to call this not knowing if it is in a browser or node environment.

PR #392, opened 2026-05-06
Audit default positions and types, add fixture, fix bugs

### LINKED ISSUE #372: Enum associated default values do not work
If I try to use the following Rust enum:

```rust
#[derive(uniffi::Enum)]
pub enum TestCase {
    One {
        #[uniffi(default = 100)]
        num_value: i32,
    },
}
```

I get the following error with uniffi-bindgen-react-native when building for WASM:

```
SyntaxError: A property signature cannot have an initializer.
```

Which makes sense when I look at the generated source:

```typescript
constructor(inner: { numValue: /*i32*/number = 100 }) {
  super("TestCase", "One");
  this.inner = Object.freeze(inner);
}
```

I [started work in my own fork for a solution](https://github.com/alastaircoote/uniffi-bindgen-react-native/commit/eda0e4ea72e440af569f687b34762b61d6a0badd) to this that instead generates an object with an optional type and code to check:

```typescript
constructor(inner: { numValue?: /*i32*/ number }) {
  super('TestCase', 'One');
  inner['numValue'] = inner['numValue'] ?? 100;

  this.inner = Object.freeze(inner);
}
```
I'm not able to continue the work right now (turns out I can't use defaults for other reasons) but wanted to provide it as a starting point for someone.

jhugman:
Fixes #372

This PR adds a defaults fixture to exercise each position and type where defaults can appear with uniffi's procmacros.

We fix three bugs:
- enum variants with structs
- enum variants with tuples
- `T::default()` in arg positions.

Thanks @alastaircoote for filing the issue!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2034,6 +2034,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "uniffi-fixture-defaults"
+version = "0.22.0"
+dependencies = [
+ "ubrn_fixture_testing",
+ "ubrn_macros",
+ "uniffi",
+]
+
+[[package]]
 name = "uniffi-fixture-enum-types"
 version = "0.22.0"
 dependencies = [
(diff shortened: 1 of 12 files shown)


zzorba (approved):
LGTM

PR #393, opened 2026-05-08
Suffix JsCallbackFn idents with module name

markharding:
## What                                                                                    
                                                                                             
  All `ForeignFutureCompleteX` callback structs are declared with the same Rust ident        
  `JsCallbackFn` in different modules. [wasm-bindgen 0.2.114+](https://github.com/wasm-bindgen/wasm-bindgen/pull/4977) collapses same-ident
  `#[wasm_bindgen]` structs into one JS class, so every `.call()` ends up dispatching to the 
  same wasm function (the last one defined — `foreignfuturecompletevoid_call` in our case).

  Reading a `ForeignFutureResult<RustBuffer>` (56 bytes) out of a void result (32 bytes)     
  picks up out-of-bounds memory. `call_status.code` lands on the low byte of a dangling
  pointer (= 1 = `Error`), so uniffi calls `lift_error` on a garbage `RustBuffer` and panics:
                  
  panicked at uniffi_core/src/ffi/rustbuffer.rs:174:                                         
  buffer capacity negative or overflowed: TryFromIntError(())
                                                                                             
  In the browser this shows up as a bare `RuntimeError: unreachable`.
                                                                                             
  ## Fix          
                                                                                             
  Suffix the generated ident with the module name so each callback struct is a distinct Rust 
  type. JS-facing class names (controlled by `js_name`) are unchanged.
                                                                                             
  ## Verify       

  Before:
  $ grep -c "^class ForeignFutureComplete|^export class ForeignFutureComplete"
  target/web/index.js                                                                        
  1
                                                                                             
  After:          
  $ grep -c "^class ForeignFutureComplete|^export class ForeignFutureComplete"
  target/web/index.js                                                                        
  12
                                                                                             
  ## Related      

  wasm-bindgen [#4977](https://github.com/wasm-bindgen/wasm-bindgen/pull/4977) — the change  
  that introduced the dedup behavior.

### DIFF
--- a/crates/ubrn_bindgen/src/bindings/gen_rust/mod.rs
+++ b/crates/ubrn_bindgen/src/bindings/gen_rust/mod.rs
@@ -98,8 +98,16 @@ impl FlavorParams<'_> {
     }
 }
 
-fn callback_fn_ident() -> Ident {
-    ident("JsCallbackFn")
+// wasm-bindgen >=0.2.114 collapses #[wasm_bindgen]-exposed structs that
+// share an identical Rust ident into one JS class breaking foreign-future callbacks.
+// Suffix the ident with the containing module's name so each is a distinct Rust type.
+// See https://github.com/wasm-bindgen/wasm-bindgen/pull/4977
+fn callback_fn_ident(module: &Ident) -> Ident {
+    use heck::ToUpperCamelCase;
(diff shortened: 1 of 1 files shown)


jhugman (approved):
This looks great! Thanks for your contribution!

PR #398, opened 2026-05-25
Expose require for library resolution in command line

jhugman:
Before this PR, the code generation could use `require` to resolve the binary library, but did not expose it to the bindgen command line.

This PR exposes it. 

It also adds the ability to customize how the triple iis exposed, either rust style or node style.

### DIFF
--- a/crates/ubrn_bindgen/src/bindings/gen_typescript/ffi_module_player/nodes.rs
+++ b/crates/ubrn_bindgen/src/bindings/gen_typescript/ffi_module_player/nodes.rs
@@ -15,8 +15,36 @@ pub enum LibResolution {
     Colocated,
     /// Bake an absolute path into the generated code.
     Absolute(Utf8PathBuf),
-    /// Resolve via `<base>-<triple>` platform packages.
-    Require(String),
+    /// Resolve via `<base><triple>` platform npm packages.
+    ///
+    /// `base` is the literal prefix joined to the triple — callers are
+    /// responsible for any separator (`-`, `/`, `_`); the runtime concatenates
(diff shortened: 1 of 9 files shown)


crates/ubrn_bindgen/src/bindings/gen_typescript/ffi_module_player/nodes.rs
```diff
@@ -15,8 +15,36 @@ pub enum LibResolution {
...
+
+/// Which platform-triple naming convention the consuming npm packages use.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+pub enum TripleStyle {
+    /// `aarch64-apple-darwin`, `x86_64-unknown-linux-gnu`, … (cargo `--target`).
+    #[default]
+    Cargo,
+    /// `darwin-arm64`, `linux-x64-gnu`, … (napi-rs convention).
+    Node,
+}
```
1egoman, line 38: thought: I wonder in practice how useful this will actually be. Can you outline some cases where one format would be preferred over the other that justifies the complexity of supportingboth?

1egoman (approved):

crates/ubrn_bindgen/src/bindings/gen_typescript/ffi_module_player/nodes.rs
jhugman: yes; I built this out to match how napi publishes, then realized that we were actually using Rust style triples.

I think defer the decision: shall we bless one or the other or both when we pull packaging in to UBRN.

PR #399, opened 2026-05-25
Rename typescript package to `ubjs/core`

jhugman:
This PR moves the name of the typescript runtime used in all generated code to `@ubjs/core`.

This is so as to reduce the amount of shipped in new `npm install`, and starts the process of a rename away from uniffi-bindgen-react-native.


.github/workflows/npm.yml
```diff
@@ -1,22 +1,37 @@
...
+  publish-ubrn:
     runs-on: ubuntu-latest
     permissions:
       contents: read
       id-token: write
     steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-node@v4
+      - uses: actions/checkout@v6
+      - uses: actions/setup-node@v6
```
1egoman, line 20: nitpick: you may want to consider pinning your github actions to a commit hash prevent some variants of supply chain attacks, [more info here](https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions)

1egoman (approved):

.github/workflows/npm.yml
jhugman: Interesting.

PR #401, opened 2026-05-29
Fix wasm template optimization setting for release profile

MrCreativ3001:
Updates the generated wasm template to use the correct numeric value for `opt-level` in the release profile.

Before
```toml
[profile.release]
# Tell `rustc` to optimize for speed
opt-level = "3"
```
After
```toml
[profile.release]
# Tell `rustc` to optimize for speed
opt-level = 3
```

According to the Cargo profile documentation, opt-level accepts integer values (0-3) or specific string values such as "s" and "z".

Reference:
https://doc.rust-lang.org/cargo/reference/profiles.html#opt-level


jhugman (approved):
Thanks for your contribution!

PR #407, opened 2026-07-07
Export package.json subpath

### LINKED ISSUE #404: Android CMake cannot resolve package.json because it is not exported
### Summary

`uniffi-bindgen-react-native@0.31.0-3` generates Android CMake that resolves the package root with:

```cmake
COMMAND node -p "require.resolve('uniffi-bindgen-react-native/package.json')"
```

That fails in Node environments that enforce package `exports`, because the published package exports only `"."` and does not export `"./package.json"`.

### Reproduction

In a clean temporary directory outside our workspace:

```sh
cat > package.json <<'JSON'
{
  "private": true,
  "dependencies": {
    "uniffi-bindgen-react-native": "0.31.0-3"
  }
}
JSON

pnpm install
```

Then run:

```sh
node -p "require.resolve('uniffi-bindgen-react-native/package.json')"
```

In our project this is hit during Android native build after:

```sh
ubrn build android --and-generate --config ubrn.config.yaml
```

because the generated Android `CMakeLists.txt` contains:

```cmake
execute_process(
    COMMAND node -p "require.resolve('uniffi-bindgen-react-native/package.json')"
    OUTPUT_VARIABLE UNIFFI_BINDGEN_PATH
    OUTPUT_STRIP_TRAILING_WHITESPACE
)
get_filename_component(UNIFFI_BINDGEN_PATH "${UNIFFI_BINDGEN_PATH}" DIRECTORY)
```

### Actual result

Node throws:

```text
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './package.json' is not defined by "exports" in .../node_modules/uniffi-bindgen-react-native/package.json
```

### Expected result

The generated Android CMake should be able to resolve the package root naturally, without consumers patching generated files or relying on the internal location of the package entrypoint.

### Environment

- `uniffi-bindgen-react-native`: `0.31.0-3`
- Node: `24.14.1`
- Package manager: `pnpm` `10.33.0` in the isolated reproduction
- Android build via React Native / Expo dev client
- Generated binding style: JSI / Android CMake template

This should not be pnpm-specific: the package has an `exports` map, and `./package.json` is not listed there, so modern Node resolution rejects this subpath.

### Proposed fix

Export `./package.json` from `uniffi-bindgen-react-native/package.json`:

```diff
 "exports": {
   ".": {
     "types": "./typescript/dist/esm/index.d.ts",
     "import": "./typescript/dist/esm/index.js",
     "require": "./typescript/dist/cjs/index.js"
-  }
+  },
+  "./package.json": "./package.json"
 }
```

We tested this locally via a `pnpm` package patch. After adding the export:

```sh
node -p "require.resolve('uniffi-bindgen-react-native/package.json')"
```

resolves successfully, `ubrn build android --and-generate` regenerates the original clean CMake line unchanged, and the Android app builds successfully.

### Workaround used locally

We patched the package metadata only, not generated CMake:

```diff
diff --git a/package.json b/package.json
--- a/package.json
+++ b/package.json
@@ -20,7 +20,8 @@
       "types": "./typescript/dist/esm/index.d.ts",
       "import": "./typescript/dist/esm/index.js",
       "require": "./typescript/dist/cjs/index.js"
-    }
+    },
+    "./package.json": "./package.json"
   },
```

This keeps the Android template behavior intact and avoids brittle path walking such as resolving the entrypoint and navigating upward through `typescript/dist/cjs`.

DeyLak:
## Summary

Export `./package.json` from the root package `exports` map.

## Why

The generated Android CMake resolves the package root with `require.resolve('uniffi-bindgen-react-native/package.json')`. When Node enforces package `exports`, that subpath currently fails with `ERR_PACKAGE_PATH_NOT_EXPORTED` because only `.` is exported.

## Validation

Ran:

```sh
node -p "require.resolve('uniffi-bindgen-react-native/package.json')"
git diff --check
```

The resolver now returns the package `package.json` file successfully.

Fixes #404.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,8 @@
       "types": "./typescript/dist/esm/index.d.ts",
       "import": "./typescript/dist/esm/index.js",
       "require": "./typescript/dist/cjs/index.js"
-    }
+    },
+    "./package.json": "./package.json"
   },
   "bin": {
     "ubrn": "./bin/cli.cjs",


jhugman (approved):
This looks right; I don't think it has any other implications.

jhugman:
@DeyLak Thank you so much for your contribution!

PR #408, opened 2026-07-13
Add `forceAsync` to `uniffi.toml`.

jhugman:
In preparation for the OtMT work, and to aid migration of a codebase to use OtMT, then add a `forceAsync` flag to `uniffi.toml`.

This is documented [here](https://github.com/jhugman/uniffi-bindgen-react-native/blob/8b294ca9a34e68477d1a7dc7eaaf629f39099b83/docs/src/reference/uniffi-toml.md#forcing-an-async-surface).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2130,6 +2130,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "uniffi-fixture-force-async"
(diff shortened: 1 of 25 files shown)


zzorba (approved):
LGTM

PR #409, opened 2026-07-14
update `uniffi` to 0.32 version

vivianjeng:
## Summary

Updates `uniffi` to 0.32.0 and brings the TypeScript generator's "pipeline" code path in line with the upstream API changes that version introduces.

- Bump `uniffi`, `uniffi_bindgen`, `uniffi_core`, `uniffi_meta` from 0.31.0 → 0.32.0, and pull in the associated `askama` (0.16.0) / `cargo_metadata` (0.23.1) upgrades required by the new `uniffi_bindgen`.
- Widen `ubrn_bindgen`'s `toml` requirement from `>=0.8, <=0.9` to `>=0.9, <2` so it can resolve against the `toml` version `uniffi_bindgen` 0.32.0 now pulls in — this was causing `E0308` type mismatches during `cargo install`.
- Update the TypeScript pipeline generator for uniffi 0.32's metadata changes:
  - Track enums with explicit discriminants (`collect_explicit_discr_enums` in `cli.rs`) and thread that `HashSet<String>` through `generate_api_from_pipeline` / `ApiModule::pipeline` / `from_general` so explicit-discriminant enums are handled correctly during TS codegen.
  - Add an `orig_name: None` field to `FnMetadata` construction sites, and switch `RustCrate` package-name extraction from `.clone()` to `.to_string()`.
  - Simplify FFI type access in the `ffi_module` and `ffi_module_player` builders (drop redundant `.ty` dereferences) and add `EmptySet`/`Box`/`Set` handling (with `unimplemented!` stubs for `Box`/`Set` labels) in the API module builders and type helpers.
- Add self-referencing `"./package.json": "./package.json"` export entries to the root and `typescript/` `package.json` files so consumers can resolve each package's own `package.json` via its exports map.

## Test plan

- [ ] `cargo build` / `cargo test` across the workspace with `uniffi` 0.32.0
- [ ] `cargo install --path crates/ubrn_bindgen` succeeds without the previous `E0308` toml mismatch
- [ ] Generate TypeScript bindings for a fixture crate containing an enum with an explicit discriminant and confirm the generated code is correct
- [ ] Existing integration/fixture test suite passes


crates/ubrn_bindgen/src/bindings/gen_typescript/api_module/type_helpers.rs
```diff
@@ -101,6 +101,12 @@ pub(super) fn type_label_for(config: &Config, ty: &general::Type) -> String {
             type_label_for(config, key_type),
             type_label_for(config, value_type)
         ),
+        general::Type::Box { .. } => {
+            unimplemented!("Box types are not yet supported by uniffi-bindgen-react-native")
+        }
+        general::Type::Set { .. } => {
+            unimplemented!("Set types are not yet supported by uniffi-bindgen-react-native")
+        }
```
jhugman: 🤔

crates/ubrn_bindgen/Cargo.toml
```diff
@@ -19,7 +19,7 @@ heck = { workspace = true }
 paste = { workspace = true }
 serde = { workspace = true }
 textwrap = "0.16.1"
-toml = ">=0.8, <=0.9"
```
jhugman, line 22: Hygiene: could you remove the Claude authorship of this commit, so it doesn't show up as a co-author to the PR.

crates/ubrn_bindgen/src/bindings/gen_typescript/api_module/mod.rs
```diff
@@ -379,6 +380,12 @@ impl TsApiModule {
                         defs.push(TsTypeDefinition::StringHelper(build_string_helper(flavor)));
                     }
                 }
+                general::TypeDefinition::Box(_) => {
+                    unimplemented!("Box types are not yet supported by uniffi-bindgen-react-native")
+                }
+                general::TypeDefinition::Set(_) => {
+                    unimplemented!("Set types are not yet supported by uniffi-bindgen-react-native")
+                }
```
jhugman, line 391: This should be implemented.

jhugman (changes requested):
Thanks for your PR!

I have reviewed the PR and left some remarks. 

Could you also have a look at the and make sure that we have covered all the new features in this bindgen that are mentioned in the uniffi-rs changelog?

There are definitely new features in uniffi-rs which we should be supporting in all UBRN; these may need additional tests!

vivianjeng:
Hi @jhugman please check the latest update 🙏🏻

jhugman:
Hi @vivianjeng could you have a look at [uniffi-rs's changelog](https://github.com/mozilla/uniffi-rs/blob/main/CHANGELOG.md), and the commit history for that project and see if anything more needs supporting in uniffi-bindgen-react-native.

I know there is at least recursive enums and zero copy array buffers which will need at least a test in this repo to show it works.

Thank you for your hard work.

vivianjeng:
Hi @jhugman, thanks for the pointer! I went through the uniffi-rs changelog (0.31.x → 0.32.0 and the unreleased section) and checked this repo against it.

**Recursive enums** — added an `IntList` recursive enum fixture (linked-list style, self-referential variant) with a full round-trip test in `fixtures/enum-types`.

**Zero-copy array buffers** — added a new `fixtures/array-buffers` fixture, a sibling to `strict-byte-arrays`. It runs the same round-trip suite (well-known bytes, identity round-trips via both lift/lower and forced read/write, and a size sweep from 1 byte up to 32 MB) but against the default TypeScript config, so it exercises the plain `ArrayBuffer`/`FfiConverterArrayBuffer` cursor path that `strict-byte-arrays` doesn't cover (that one forces `Uint8Array`). All three targets (JSI, WASM, Napi) pass.

One more thing I found while digging through the changelog: 0.32.0 also added zero-copy `[ByRef] bytes` arguments (`ForeignBytes`, pointer+length instead of a `RustBuffer` copy) for Kotlin/Swift/Python (#2878). We don't support that convention at all yet — `runtimes/core`'s CIF layer explicitly rejects `ForeignBytes` as a real argument type today (it's only used internally for one string helper). It's not a quick add — it'd need libffi struct-by-value marshaling plus raw-pointer extraction from JS `ArrayBuffer`s on both JSI and Napi, and there's no existing JS-bindgen reference implementation upstream to follow. Happy to scope that out separately if it's worth doing.

Nothing else in the 0.32.0 or unreleased notes looked relevant (the Kotlin JNI bindgen and `uniffi_parse_rs` items are Kotlin/tooling-specific). Thanks again for flagging this!

PR #415, opened 2026-07-22
fix(napi): support runtimes without external ArrayBuffers

injaneity:
## Summary

Electron rejects the NAPI runtime's zero-copy RustBuffer handoff because V8's memory cage disallows external ArrayBuffers, so UniFFI buffer-backed returns fail with `napi_no_external_buffers_allowed`.

- preserve the zero-copy RustBuffer path when the runtime supports external ArrayBuffers
- fall back to a copied, V8-owned `Uint8Array` when NAPI reports that external buffers are unavailable
- release the original Rust allocation immediately and mark copied views so generated `rustbuffer_free(view)` calls are safe no-ops
- apply the same fallback to `rustbuffer_alloc`, allowing generated lowering code to work in Electron
- retain non-Electron errors instead of masking them as the unsupported-runtime case

Node documents Electron as a runtime that can return `napi_no_external_buffers_allowed`: https://nodejs.org/api/n-api.html#napi_create_external_arraybuffer

## Validation

- `cargo fmt --all -- --check`
- `cargo check -p uniffi-runtime-napi`
- `cargo clippy -p uniffi-runtime-napi --all-targets -- -D warnings`
- built the NAPI test libraries and ran all 89 runtime tests
- replaced `@ubjs/node-darwin-arm64` with the patched addon in an Electron 41 host and started a real Rust-backed `EmbeddedCuaDriverHost`
- verified the Electron host received the complete connection record, propagated its MCP configuration, directly owned the daemon child, and removed the child and private socket on shutdown

The Electron validation uses the runtime's real buffer-backed async record return, which failed before this change with `Failed to create external ArrayBuffer`.

### DIFF
--- a/runtimes/napi/src/call/mod.rs
+++ b/runtimes/napi/src/call/mod.rs
@@ -172,37 +172,37 @@ pub(crate) fn call_ffi_function(
     }
 
     match &call_ret {
-        CallReturn::RustBuffer(rb) => {
-            rust_buffer_to_js_uint8array_handoff(env, *rb, capacity_symbol)
-        }
+        CallReturn::RustBuffer(rb) => rust_buffer_to_js_uint8array_handoff(
+            env,
+            *rb,
+            module.rb_ops().free_ptr,
+            capacity_symbol,
+        ),
         _ => marshal::read_return_to_js(env, &call_ret),
     }
 }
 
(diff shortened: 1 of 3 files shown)


injaneity:
closing at the author’s request. the validated patch remains available locally while we coordinate the dependency fix through the consuming project.

PR #417, opened 2026-07-22
fix(napi): support runtimes without external ArrayBuffers

injaneity:
## Summary

Electron's V8 memory cage rejects external ArrayBuffers, so RustBuffer returns fail with `napi_no_external_buffers_allowed`.

- keep the zero-copy return path where external buffers are supported; otherwise copy into V8 memory and release the Rust allocation immediately
- keep lowering buffers JS-owned because call dispatch already copies them into the RustBuffer consumed by FFI
- track ownership with non-enumerable metadata so cleanup is idempotent, unowned V8 pointers never reach Rust's allocator, and empty returns release spare Rust capacity
- add focused ownership tests and an Electron regression to N-API CI

This is the fallback recommended by the [N-API documentation](https://nodejs.org/api/n-api.html#napi_create_external_arraybuffer) for runtimes that disallow external buffers.

## Validation

- rustfmt and clippy with warnings denied
- all 90 N-API runtime tests
- Electron 39 regression test
- Electron 41 `EmbeddedCuaDriverHost` lifecycle, including connection decoding, direct child ownership, and process/socket cleanup

### DIFF
--- a/runtimes/napi/src/register/mod.rs
+++ b/runtimes/napi/src/register/mod.rs
@@ -38,12 +38,9 @@ pub fn register(
     let functions: JsObject = definitions.get_named_property("functions")?;
     let mut result = env.create_object()?;
 
-    // Per-registration Symbol used as a hidden capacity-hint key on lift-
-    // handoff `Uint8Array` views. The view-handoff path returns a view whose
-    // `byteLength` is `rb.len`, but Rust may have allocated
-    // `rb.capacity > rb.len`. We stash `capacity` on the view via this
-    // symbol; `rustbuffer_free(view)` reads it back when releasing the
-    // allocation.
+    // Per-registration Symbol used as hidden ownership metadata on RustBuffer
+    // views. Rust-owned lift views carry their allocation capacity; JS-owned
+    // views carry zero so free is a no-op.
     //
     // SAFETY: env is the active napi env supplied by node for this register
     // call. The `Arc<CapacitySymbol>` keeps the Symbol alive across the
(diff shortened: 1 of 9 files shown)


runtimes/napi/tests/electron-rust-buffer.cjs
```diff
@@ -0,0 +1,52 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+const assert = require("node:assert/strict");
+const { app } = require("electron");
```
jhugman, line 7: Could this test file run all the tests, just under electron?

runtimes/napi/src/register/mod.rs
```diff
@@ -125,38 +111,33 @@ pub fn register(
...
-        //       the stashed value via `CapacitySymbol::get`. When
-        //       `capacity == len` at handoff time we skip the property write,
-        //       so absence of a hint here is also a valid "use byteLength"
-        //       signal.
-        // SAFETY: raw_env / raw_val are valid for the current callback scope.
-        let capacity = unsafe { cap_sym_for_free.get(raw_env, raw_val) }.unwrap_or(length as u64);
+
+        // Mark the view released before handing its original allocation back to
+        // Rust. Repeated cleanup is then a no-op instead of a double free.
+        unsafe { cap_sym_for_free.set(raw_env, raw_val, 0)? };
```
jhugman, line 133: Could you add [Safety comments](https://std-dev-guide.rust-lang.org/policy/safety-comments.html) for each of the unsafe blocks you've introduced?

jhugman (commented):
@injaneity thanks so much for this contribution!

I have a couple of comments added, mainly around re-using tests, and safety comments.

Also: could you add something in to the `CHANGELOG.md`, taking credit for yourself.

PR #418, opened 2026-07-23
feat(wasm2): add the player runtime

jhugman:
Node and browser entries differ only in how they fetch the module, so the
core stays environment-neutral and bundles anywhere.

Add `uniffi-runtime-wasm`, the helper crate a consuming cdylib links. It
exports the allocator pair the player allocates through and forwards Rust
panics to a JS logger installed in the function table.

Exclude `runtimes/wasm` from the workspace glob, since it holds the npm
package rather than a crate, and register the helper crate as a member.

Note in the napi runtime that its capacity symbol has a wasm2 counterpart.

### DIFF
--- a/runtimes/wasm/core/src/callback.ts
+++ b/runtimes/wasm/core/src/callback.ts
@@ -1,0 +1,725 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/
+ */
+// JS-side dispatch table and per-shape trampoline cache for the Wasm2 player.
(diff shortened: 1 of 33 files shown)


runtimes/wasm/core/src/callback.ts
```diff
@@ -0,0 +1,725 @@
...
+          case "RustBuffer":
+            writeRustBufferPayload(
+              memory,
+              alloc,
+              ptr + f.offset,
+              fieldVal as Uint8Array,
+            );
+            break;
+          default:
+            f.shape.write(memory, ptr + f.offset, fieldVal);
```
zzorba, line 183: Is this an intentional fall through?  Maybe worth a comment?

zzorba (approved):
LGTM

PR #424, opened 2026-08-03
feat(common): add wasm post-processing and crate queries

jhugman:
Add a `wasm` feature carrying the post-link steps the player needs on a
`wasm32-unknown-unknown` cdylib:

  * `stage_wasm` copies the module beside the generated TypeScript, running
    wasm-bindgen's rewriter first when the module imports its placeholder
    namespace.
  * `export_growable_table` exports the function table with no upper bound,
    so the player can grow it to install callback trampolines. Rewriting
    post-link avoids forcing RUSTFLAGS on every consumer.
  * `dce_wasm` strips exports the player cannot reach, then runs walrus' gc.

Add `RustCrate` queries for the manifest facts the wasm2 build validates:
whether the crate is a cdylib, and whether it declares a given dependency or
feature. Read the declared edge rather than the resolved graph, because
`cargo metadata` on a workspace member unions features across every member.

Add `run_in_process`, so work done through a library rather than a
subprocess still shows up in the command-recording test harness.

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,18 @@
 version = 4
 
 [[package]]
+name = "ahash"
+version = "0.8.12"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5a15f179cd60c4584b8a8c596927aadc462e27f2ca70c04e0071964a73ba7a75"
+dependencies = [
+ "cfg-if",
+ "once_cell",
+ "version_check",
+ "zerocopy",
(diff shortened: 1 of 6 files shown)


zzorba (approved):
LGTM

PR #432, opened 2026-08-07
Adopt library-owned RustBuffer arguments instead of copying them to attempt to address napi player memory leak

1egoman:
### Problem

After some benchmarking, I have determined that the NAPI player runtime leaks memory for each FFI call that has a `RustBuffer` argument.

My use case: a 48 kHz audio plugin sends one 1924-byte buffer for each 10 ms frame. The plugin therefore leaks approximately 2.4 KB for each frame, which equals approximately 14.7 MB for each minute of audio - fairly signifigant.

### Cause

The runtime allocates each argument buffer two times. The runtime releases only one of the two buffers.

The generated code lowers a `RustBuffer` argument in three steps:

1. The generated code calls `rustbuffer_alloc(n)`. The library allocates buffer A, and the runtime gives JavaScript an external `Uint8Array` view of buffer A.
2. The generated code writes the value into buffer A.
3. The generated code sends the view as the argument.

The argument then goes to `js_uint8array_to_rust_buffer`. That function called `rustbuffer_from_bytes`, which allocated buffer B and copied buffer A into buffer B. The callee then released buffer B.

No code released buffer A. The generated code releases returned buffers, but it does not release lowered arguments. Also, the view has a no-op finalizer, therefore buffer A stayed in memory until the process stopped.

The `rustbuffer_alloc` and `rustbuffer_free` pair exists to prevent return direction already used the pair correctly. Only the argument direction did not.

### Solution

> [!WARNING]
> I'm not fully confident in this solution, and I used a fair bit of LLM help to get to it. In particular, using `capacity` as a hint like this to me seems like it could be quite fragile, but I also am not sure where this metadata would live otherwise other than adding a whole new level of abstraction around `RustBuffer`. Definitely open to suggestions and alternate approaches! 

The runtime now adopts library-owned buffers - The callee receives the existing allocation, and the runtime does not copy the buffer.

Two kinds of view come to this function. The runtime must treat them differently:

- Library-owned views come from rustbuffer_alloc. The runtime adopts these views.
- V8-owned views are ordinary `Uint8Array` objects from JavaScript. The runtime must copy these views and must not give V8 memory to the library.

The capacity hint symbol is the marker. The runtime knows a buffer's capacity only when the library allocated the buffer. Therefore, a capacity hint is present if and only if the library owns the memory.

To keep this invariant, `rustbuffer_alloc` now stamps the capacity on each view that it returns.

After adoption, the runtime sets the hint to 0. This prevents a double release. The `free_rustbuffer` function ignores buffers that have a capacity of 0. Therefore a later `rustbuffer_free(view)` call does nothing.

If a caller sends an adopted view a second time, the runtime returns an error, as the callee has already released it.

### Test

The new tests count the live allocations in the fixture. They do not sample RSS. V8 can change its own heap reservation by several megabytes, which hides a leak of a few hundred bytes for each call. A counter is used to give an exact equality instead.

List of test cases:

1. A lowered argument buffer is released.
2. A lowered argument buffer is released for a call that return
3. Both argument buffers are released for a call that has 2 arguments.
4. A plain JavaScript Uint8Array is copied and not adopted.
5. An alloc and free pair is balanced.
6. A release of an adopted view does not cause a double release
7. An empty view is correct.

Before the fix, tests 1, 2, and 3 failed. Test 1 reported leaked 64 buffers over 64 calls. This is one buffer for each call.

After the fix, all 7 tests pass.

### Limits of this change

Two items remain. Neither is a new problem, and neither is a result of this change.

1. The callback paths still copy. No capacity symbol is available in those paths. To adopt buffers there, a person must send the symbol through the callback user data and the vtable marshalling code. The copy is always safe, but a rustbuffer_alloc view that a callback lowers is still orphaned.
2. A second, independent leak remains. Each external `Uint8Array` view leaks approximately 193 bytes. This quantity does not change with the buffer size. The alloc/free test measures 194.6 bytes for each cycle both before and after this change. The cause is in `create_external_uint8array`, and not in buffer ownership. This item needs a separate investigation.

### DIFF
--- a/crates/ubrn_bindgen/src/bindings/gen_cpp/templates/RustBufferHelper.cpp
+++ b/crates/ubrn_bindgen/src/bindings/gen_cpp/templates/RustBufferHelper.cpp
@@ -30,18 +30,58 @@ template <> struct Bridging<RustBuffer> {
   static RustBuffer fromJs(jsi::Runtime &rt, std::shared_ptr<CallInvoker>,
                            const jsi::Value &value) {
     try {
+      auto obj = value.asObject(rt);
+
+      // Adoption vs copy. Two kinds of view arrive here:
+      //
+      //   * Library-owned views from `rustbuffer_alloc`. Codegen allocates one,
+      //     fills it in place, and ships it as an argument — and never frees a
+      //     lowered argument, while the view's backing `CMutableBuffer` is
+      //     non-owning (its destructor leaves `data` alone). Such a view carries
+      //     a capacity hint (stamped by `rustbuffer_alloc` in the wrapper). We
+      //     *adopt* it: hand the existing allocation to the callee, which frees
+      //     it. Copying instead would orphan the allocation and leak one whole
+      //     payload per call.
+      //   * Ordinary JS-owned arrays, which carry no hint. These are *copied*
+      //     into a fresh library allocation — they are not ours to give away.
+      //
+      // On adoption we reset the hint to 0 so a later `rustbuffer_free(view)` is
+      // a no-op rather than a double free.
+      if (obj.hasProperty(rt, uniffi_jsi::kUbrnRustCapacity)) {
+        auto capacity = static_cast<uint64_t>(
+            obj.getProperty(rt, uniffi_jsi::kUbrnRustCapacity).asNumber());
+        if (capacity == 0) {
+          // Hint present but zeroed: the view was already adopted by a previous
+          // call and its memory has been freed. Reading it would be a
+          // use-after-free, so refuse rather than hand over a dangling pointer.
+          throw jsi::JSError(
+              rt,
+              "RustBuffer argument was already consumed by a previous FFI call");
+        }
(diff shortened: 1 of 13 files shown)


runtimes/napi/package-lock.json
```diff
@@ -1,12 +1,12 @@
...
   "name": "@ubjs/node",
-  "version": "0.31.0-2",
+  "version": "0.31.0-3",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@ubjs/node",
-      "version": "0.31.0-2",
+      "version": "0.31.0-3",
```
1egoman, line 9: Note - the version number was out of date in the `package-lock.json`, so I fixed it here. I can drop this commit if you'd prefer to fix this out of band.

jhugman (approved):
I added a fix for the latent JSI bug that this reveals.

Issue #412, opened 2026-07-15
OtMT: Off the Main Thread — Channels proposal


jhugman:
The approach to do next is getting the Rust off the main thread. This will increase the overhead per call, but allow for longer running Rust calls which would otherwise block the UI.

This will open up many opportunities for improving apps built with uniffi, ubrn and Rust e.g.

- Targeting SharedWorkers as well as WebWorkers
- Electron apps running real Rust, instead of WASM.

Currently, high level data types are _lowered_ into FFI representations: mostly `Uint8Array`s, then called into Rust via a "player". The lowering happens in generated API, in Typescript. The player is varied depending on what context the JS is running in: React Native has a JSI player, Node has a NAPI player, and WASM has a WASM player.

Players are configured by typescript at initialization, and use (mostly) dlopen and libffi to dynamically dispatch calls into Rust. They open the Rust binary, and return a Typescript object with generated methods that can call the low-level C ABI generated by uniffi-rs. These are entirely inspired by [ffi-rs][ffi-rs].

[ffi-rs]: https://www.npmjs.com/package/ffi-rs

The plan is to add a new concept, which I'm calling a Channel, which is a remote/proxy interface onto a player. It has a `ui` side and a `worker` side. From the UI side, the generated API uses a similar object that the configured player provides: the same except that _all_ calls into and out of Rust are asynchronous.

The channel is an abstraction over an IPC channel, between two JS Contexts. We keep the interfaces similar to the players, so we can adapt the players that don't support a second JS context (e.g. JSI and NAPI).

We can mix and match the channels & players to give us real choices:

- a WASM player: just the same as we have now. When responsiveness becomes a problem, migrate to…
- a WebWorker channel + WASM player. This might work for the web, and a fast path to electron support.
- an Electron IPC channel + NAPI player would give you real Rust on the desktop.
- a WASM player, with a WASM channel might have much of the API on the ui side, but also an API for hand written worker code
- a WebWorker channel, with some small changes, may be support SharedWorkers.

## Changes to bookkeeping

Some ui side bookkeeping remains the same: 
- handle maps for objects and callback interfaces
- finalization registries

Some ui side bookkeeping moves closer to the player:
- unresolved Rust futures implemented as Promises

A new class of bookkeeping is needed by the channels:
- a mapping between calls and their returns
- additional lifecycle methods 

## Phases

[...]

PR #433, opened 2026-08-07
fix(napi): answer is_js_thread per callback, not per process

### LINKED ISSUE #436: napi: a second JS thread deadlocks on its first async call
**What I was doing**

Calling a uniffi library from a Node `worker_threads` worker. Vitest and Jest both default to a worker pool, so this also happens to people who never asked for threads.

**What happens**

The first call from that worker never returns. The thread is parked in native code, so `worker.terminate()` never resolves and `process.exit()` never takes effect; the process has to be killed from outside.

It isn't limited to calls with a callback of your own. A uniffi future's continuation is passed as an `FfiType.Callback` argument, so *every* async call from a second JS thread hangs. Synchronous exports still work, which makes it easy to miss: a smoke test can pass while the real workload wedges.

**Why**

```mermaid
sequenceDiagram
    participant Main as Main thread
    participant Worker as Worker thread
    participant Rust

    Main->>Rust: loads the addon
    Note over Rust: module_init runs once per library load<br/>MAIN_THREAD_ID = Main thread

    Worker->>Rust: calls a function with a callback
    Rust->>Rust: is_js_thread()
    Note over Rust: current is Worker,<br/>MAIN_THREAD_ID is Main,<br/>so: not a JS thread
    Rust-->>Worker: queues the callback on the Worker's own event loop
    Rust->>Worker: blocks the Worker until that callback replies
    Note over Worker: blocked, so it never runs its own queue,<br/>so the callback never replies
```

`is_js_thread` answers from `MAIN_THREAD_ID`, a `OnceLock<ThreadId>` set in `#[module_init]`. That is a ctor: it runs once per library load, on whichever thread loads the addon first. Every JS thread except that one is told it is not a JS thread, takes the cross-thread path, and is made to wait for itself.

`dispatch_to_js_thread` says as much in its own safety comment: "Must NOT be called on the main thread (the blocking recv would deadlock)." A second JS thread is exactly that case, under a different name.

**Reproducing**

Register on the main thread, then invoke any callback from a worker. `runtimes/napi/tests/helpers/second-thread-callback.mjs` in #433 does it in about forty lines. Reproduced on macOS arm64 and on Debian bookworm arm64 (glibc 2.36, Node 22).

**Context**

I have a patch in #433: `CallbackUserData` already stores the registering thread's `napi_env`, its `napi_ref` and a `ThreadsafeFunction` bound to that thread's loop, so recording the `ThreadId` alongside them makes the predicate per callback rather than per process. Filing this separately so the bug is searchable on its own.

This is not the same as #412. That proxies a player from one JS context so a long call can leave the UI thread; this is the NAPI player being incorrect when several contexts use it directly, which is what a server-side Node process does when it reaches for a worker pool.

stupside:
Fixes #436. Stacked on #435: the first commit here is that PR's, so review or land this one second. It has to be underneath rather than alongside — see **Ordering** below.

**What I was doing**

Calling a uniffi library from a Node `worker_threads` worker. Vitest and Jest both default to a worker pool, so this also happens to people who never asked for threads.

**What happens**

The first call from that worker never returns. The thread is parked in native code, so `worker.terminate()` never resolves and `process.exit()` never takes effect; the process has to be killed from outside.

It isn't limited to calls with a callback of your own. A uniffi future's continuation is passed as an `FfiType.Callback` argument, so *every* async call from a second JS thread hangs.

**Why**

```mermaid
sequenceDiagram
    participant Main as Main thread
    participant Worker as Worker thread
    participant Rust

    Main->>Rust: loads the addon
    Note over Rust: module_init runs once per library load<br/>MAIN_THREAD_ID = Main thread

    Worker->>Rust: calls a function with a callback
    Rust->>Rust: is_js_thread()
    Note over Rust: current is Worker,<br/>MAIN_THREAD_ID is Main,<br/>so: not a JS thread
    Rust-->>Worker: queues the callback on the Worker's own event loop
    Rust->>Worker: blocks the Worker until that callback replies
    Note over Worker: blocked, so it never runs its own queue,<br/>so the callback never replies
```

The check is process-wide, but what it decides is per callback. Every JS thread except the first is told it is not a JS thread, and is then made to wait for itself.

**The fix**

`CallbackUserData` already stores the registering thread's `napi_env`, its `napi_ref` and a `ThreadsafeFunction` bound to that thread's loop. Only the `ThreadId` was missing:

```diff
  is_js_thread(user_data)
-     current == MAIN_THREAD_ID           // process-wide
+     current == user_data.owner_thread   // per callback
```

`runtimes/core` already passes `user_data` into `is_js_thread`, and this implementation was discarding it, so nothing outside `runtimes/napi` changes. `MAIN_THREAD_ID` and `is_main_thread` become unused, and with them gone `#[module_init]` has nothing left to set: the source is a net eleven lines shorter.

**Test**

Registers on the main thread, then invokes a callback from a worker. It runs in a child process on purpose, because a regression parks a JS thread and would wedge the suite rather than fail it.

Checked both ways by rebuilding the addon from the parent commit: 30s deadlock without the change, 48ms with it. Suite is 91/91.

**Ordering**

This test is the first thing in the suite to create a `ThreadsafeFunction` on a second environment, which is precisely the case #434 gets wrong: only the first environment installs a cleanup hook, but every environment's handles go into one shared list. The worker's teardown destroys its TSFN, and at process exit the main environment's hook aborts the freed handle. The child dies with SIGABRT and prints nothing, so the test reports `exitCode` `null` rather than a failed assertion — deterministic on Linux, invisible on macOS.

So this PR is stacked on #435 rather than the other way round. The two bugs are independent; only the order is.

**Not in this PR**

* This is not the Channel work in #412, and does not overlap it. That proxies a player from one JS context so a long call can leave the UI thread; this makes the NAPI player itself correct when several contexts use it directly. A Node server has no UI thread to get off, and reaches for a worker pool for throughput, so being directly usable from several threads is what that case needs.
* This does not make a cdylib safe across several JS realms in general: UniFFI registers one foreign-callback vtable per cdylib (mozilla/uniffi-rs#2295). This fixes the hang; that is what remains.

### DIFF
--- a/runtimes/napi/tests/threading.test.mjs
+++ b/runtimes/napi/tests/threading.test.mjs
@@ -103,6 +103,41 @@ test("callback: receives RustBuffer arg from another thread", async () => {
   assert.deepStrictEqual(result.data, new Uint8Array([0xca, 0xfe, 0xba, 0xbe]));
 });
 
+// A second JS thread can use the runtime: it registers a callback and Rust invokes it.
+//
+// Run in a child process on purpose. A regression parks a JS thread inside native code, where
+// `worker.terminate()` never resolves and `process.exit()` never takes effect, so an in-process
+// version would wedge the suite instead of failing it.
+test("threads: a second JS thread can invoke a callback", async () => {
+  const script = join(
+    import.meta.dirname,
+    "helpers",
+    "second-thread-callback.mjs",
+  );
+  const child = spawn(process.execPath, [script], {
+    stdio: ["ignore", "pipe", "inherit"],
+  });
+
+  let out = "";
+  child.stdout.on("data", (chunk) => {
+    out += String(chunk);
+  });
+
+  const exitCode = await new Promise((resolve) => {
+    const timer = setTimeout(() => {
+      child.kill("SIGKILL");
+      resolve(null);
+    }, 30_000);
+    child.on("close", (code) => {
+      clearTimeout(timer);
+      resolve(code);
+    });
+  });
+
+  assert.match(out, /^CALLED 7 42$/m, "the worker's callback did not run");
+  assert.strictEqual(exitCode, 0, "the child did not exit on its own");
+});
+
 // One environment's teardown does not disturb another's.
 //
 // The worker registers first, so under a single process-wide cleanup hook it is the environment
--- a/runtimes/napi/tests/threading.test.mjs
+++ b/runtimes/napi/tests/threading.test.mjs
@@ -4,6 +4,8 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/
  */
 import { test } from "node:test";
+import { spawn } from "node:child_process";
+import { join } from "node:path";
 import assert from "node:assert";
 import lib from "../lib.js";
 const { UniffiNativeModule, FfiType } = lib;
@@ -99,4 +101,36 @@ test("callback: receives RustBuffer arg from another thread", async () => {
   assert.strictEqual(result.handle, 99n);
   assert.ok(result.data instanceof Uint8Array);
   assert.deepStrictEqual(result.data, new Uint8Array([0xca, 0xfe, 0xba, 0xbe]));
+});
+
+// One environment's teardown does not disturb another's.
+//
+// The worker registers first, so under a single process-wide cleanup hook it is the environment
+// that installs it, and its teardown aborts every environment's ThreadsafeFunctions and marks
+// every environment as shutting down. The main thread's callbacks then bail out early and zero
+// their return buffers, which looks like a callback that silently never ran.
+test("threads: one environment's teardown leaves another's callbacks working", async () => {
+  const script = join(import.meta.dirname, "helpers", "env-teardown.mjs");
+  const child = spawn(process.execPath, [script], {
+    stdio: ["ignore", "pipe", "inherit"],
+  });
+
+  let out = "";
+  child.stdout.on("data", (chunk) => {
+    out += String(chunk);
+  });
+
+  const exitCode = await new Promise((resolve) => {
+    const timer = setTimeout(() => {
+      child.kill("SIGKILL");
+      resolve(null);
+    }, 30_000);
+    child.on("close", (code) => {
+      clearTimeout(timer);
+      resolve(code);
+    });
+  });
+
+  assert.match(out, /^CALLED 7 42$/m, "the main thread's callback did not run");
+  assert.strictEqual(exitCode, 0, "the child did not exit on its own");
 });
--- a/runtimes/napi/src/callback/mod.rs
+++ b/runtimes/napi/src/callback/mod.rs
@@ -74,6 +74,9 @@ struct RustCallStatusForVTable {
 /// invocation and passes it through to `on_js_thread`, `dispatch_to_js_thread`,
 /// and `is_js_thread`.
 struct CallbackUserData {
+    /// State of the environment this callback belongs to, held here so an invocation can check for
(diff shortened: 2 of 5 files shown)


stupside:
There is a follow-up stacked on this one: #435, which fixes #434. Both came out of the same look at per-process state in the N-API runtime, but they are independent bugs and this PR stands on its own.

runtimes/napi/tests/threading.test.mjs
```diff
@@ -100,3 +102,38 @@ test("callback: receives RustBuffer arg from another thread", async () => {
...
+      resolve(null);
+    }, 30_000);
+    child.on("exit", (code) => {
+      clearTimeout(timer);
+      resolve(code);
+    });
+  });
+
+  assert.match(out, /^CALLED 7 42$/m, "the worker's callback did not run");
+  assert.strictEqual(exitCode, 0, "the child did not exit on its own");
```
jhugman, line 138: This test is failing; I think the fix is in #435.

runtimes/napi/tests/threading.test.mjs
```diff
@@ -100,3 +102,38 @@ test("callback: receives RustBuffer arg from another thread", async () => {
...
+  child.stdout.on("data", (chunk) => {
+    out += String(chunk);
+  });
+
+  const exitCode = await new Promise((resolve) => {
+    const timer = setTimeout(() => {
+      child.kill("SIGKILL");
+      resolve(null);
+    }, 30_000);
+    child.on("exit", (code) => {
```
jhugman: I'm not entirely sure, but:

```suggestion
    child.on("close", (code) => {
```

would close after the exit, so this would be safer. It's test code, so I'm not hugely concerned, but it's one fewer intermittent failures to worry about.

jhugman (commented):
This is excellent. 

I think we'll need to think about the split between this PR and #435: the last test in`threading.test.mjs` is failing, but the fix seems to be in #435. Would you mind having a look.

Otherwise, this is lovely. Thank you so much.

stupside:
> This is excellent.
> 
> I think we'll need to think about the split between this PR and #435: the last test in`threading.test.mjs` is failing, but the fix seems to be in #435. Would you mind having a look.
> 
> Otherwise, this is lovely. Thank you so much.

@jhugman Happy to contribute, great work on this project, #435 fixes but the split is not clean, I rebased the branches to make it cleaner. This branch has the commit from #435 so it can go green, you can check it out first.

runtimes/napi/tests/threading.test.mjs
stupside: Taken, thanks. Applied to the env-teardown test in #435 as well, which spawns a child the same way: `exit` can fire before the piped stdout has been drained, and both tests assert on what the child printed before they check its exit code.

runtimes/napi/tests/threading.test.mjs
stupside: Confirmed, and it is #434 rather than anything in this PR. This test is the first thing in the suite to create a `ThreadsafeFunction` on a second environment, which is the case that bug gets wrong: the worker's teardown destroys its TSFN, and at process exit the main environment's hook aborts the already-freed handle from the shared list. The child dies with SIGABRT and prints nothing, so what fails is `exitCode === 0` getting `null`, not the callback assertion above it.

jhugman (approved):
This is spot on, thank you.

I'll land it shortly.

stupside:
> This is spot on, thank you.
> 
> I'll land it shortly.

Thank you ! This will unlock my SDK release, when do you think this fix will land on npm ?

PR #435, opened 2026-08-07
fix(napi): give each Node environment its own shutdown state

### LINKED ISSUE #434: napi: shutdown state and the env cleanup hook are process-global
**What I was doing**

Using the N-API runtime from more than one Node environment: a `worker_threads` worker as well as the main thread. A test runner's pool does the same thing without being asked.

**What happens**

Two things, both at teardown.

A worker that registers after the main thread gets **no cleanup hook at all**, so when that worker is torn down its `ThreadsafeFunction`s are never aborted. That is the case the hook exists for; the comment on `ModuleState` says aborting them "unblocks any Rust worker threads that are waiting on `rx.recv()`" and "prevents SIGSEGV crashes on Linux where the shared library would otherwise be `dlclose()`'d while Rust threads still hold function pointers into it".

And whichever environment tears down **first** aborts every other environment's threadsafe functions, and makes `is_shutting_down()` true for callbacks belonging to environments that are still running. Those callbacks then bail out early and zero their return buffers.

**Why**

Both pieces of state are process-global, but what they describe is per environment.

```rust
fn install_env_cleanup_hook(env: &Env) {
    static INSTALLED: AtomicBool = AtomicBool::new(false);
    if INSTALLED.compare_exchange(...).is_ok() {   // only the FIRST env ever gets here
        napi_add_env_cleanup_hook(env.raw(), Some(env_cleanup_hook), null_mut())
    }
}
```

`MODULE_STATE` has the same shape: one `tsfn_handles` Vec collecting the handles of every environment, and one `shutdown` flag shared by all of them.

**Suggested fix**

Key both on the `napi_env`: give each environment its own state, install a hook per environment, and have `env_cleanup_hook` abort only that environment's handles.

The two callers of `is_shutting_down()` are both in `callback/mod.rs` and both already hold the `user_data` pointer, which carries `raw_env`, so the lookup can happen there rather than through a global. Holding an `Arc` to the per-env state on `CallbackUserData` keeps the hot path lock-free.

**Context**

Noticed while fixing the `is_js_thread` deadlock in #433, which is deliberately scoped to that one predicate and does not touch this. Happy to send a patch if the approach above sounds right.

stupside:
Fixes #434. #433 is stacked on this one, so this is the one that lands first.

**What I was doing**

Using the runtime from more than one Node environment: a `worker_threads` worker as well as the main thread. A test runner's pool does the same thing without being asked.

**What happens**

Two things, both at teardown.

A worker that registers after the main thread gets no cleanup hook at all, so when that worker is torn down its `ThreadsafeFunction`s are never aborted. That is the case the hook exists for: the comment on `ModuleState` says aborting them "unblocks any Rust worker threads that are waiting on `rx.recv()`" and "prevents SIGSEGV crashes on Linux where the shared library would otherwise be `dlclose()`'d while Rust threads still hold function pointers into it".

And whichever environment tears down first aborts every *other* environment's threadsafe functions, and makes `is_shutting_down()` true for callbacks belonging to environments that are still running. Those callbacks bail out early and zero their return buffers, so from JS the callback simply never runs and nothing says why.

**Why**

Both pieces of state are process-global, but what they describe is per environment.

```rust
fn install_env_cleanup_hook(env: &Env) {
    static INSTALLED: AtomicBool = AtomicBool::new(false);
    if INSTALLED.compare_exchange(...).is_ok() {   // only the FIRST env ever gets here
        napi_add_env_cleanup_hook(env.raw(), Some(env_cleanup_hook), null_mut())
    }
}
```

`MODULE_STATE` has the same shape: one `tsfn_handles` Vec collecting every environment's handles, and one `shutdown` flag shared by all of them.

**The fix**

Key both on the `napi_env`. Each environment gets its own state and installs its own hook, and the hook takes the raw env as its data pointer so it knows whose handles to abort:

```diff
- static INSTALLED: AtomicBool                       // one hook for the whole process
- static MODULE_STATE: OnceLock<ModuleState>         // one shutdown flag, one handle list
+ static ENV_STATES: OnceLock<Mutex<HashMap<usize, Arc<EnvState>>>>   // keyed by napi_env

- unsafe extern "C" fn env_cleanup_hook(_data)       // aborts every env's handles
+ unsafe extern "C" fn env_cleanup_hook(data)        // aborts only this env's handles
```

`CallbackUserData` holds an `Arc` to its own environment's state, so an invocation reads the shutdown flag straight out of its user data and the hot path takes no lock; the map is touched only when a callback is created and when an environment goes away.

**Test**

A worker registers and is torn down, then the main thread registers and invokes a callback. Under one process-wide hook the worker is the environment that installs it, so its teardown silences the main thread and the callback never runs.

Verified both ways by rebuilding the addon from the parent commit, and on Linux as well as macOS: passes with the change, and on the parent the callback silently never runs. Suite is 90/90 on macOS arm64 and on Debian bookworm arm64 (glibc 2.36, Node 22); `cargo test -- napi::` is 46/46.

**Why this is first in the stack**

The two bugs are independent, but the order is not. #433's regression test invokes a callback from a worker, which makes it the first thing in the suite to create a `ThreadsafeFunction` on a second environment — exactly the case that trips this bug. The worker's teardown destroys that TSFN, and at process exit the main environment's hook walks the shared list and calls `napi_release_threadsafe_function(..., abort)` on the freed handle. The child dies with SIGABRT and no output, so the test reports `exitCode` `null` rather than an assertion failure. That is deterministic on Linux and does not reproduce on macOS.

**Registration and the sweep are one decision**

`tsfns: Mutex<Option<Vec<TsfnHandle>>>` rather than a flag beside a list, where `None` means the environment is gone. Registering and sweeping are then settled under one lock, and a handle cannot be pushed after the sweep has already walked past it. One that arrives late is aborted by `register_tsfn` itself, so callers keep a single rule instead of a result they have to remember to act on.

Not a live bug, and worth saying so: registration and teardown are serialised today because both run on the environment's own JS thread. `create_callback_user_data` takes `&Env`, valid only there, and `napi_add_env_cleanup_hook` fires on that same thread, so the two cannot interleave. Nothing states that invariant and nothing enforces it; this does not lean on it, and not leaning on it costs a `match` and a comment.

Three smaller things in the same area:

- The sweep takes the handles under the lock and aborts them without it. Aborting drops pending calls and their `SyncSender`s, which is what unblocks Rust worker threads, and those should not be able to arrive on a lock the sweep is still holding.
- Poisoned locks are recovered rather than `expect`ed. A panic raised inside a napi callback is an abort, and this guards a list of opaque handles with no cross-field invariant a panic could break.
- `unsafe impl Send`/`Sync` moved off `EnvState` onto a `TsfnHandle` newtype, so the safety comment sits on the raw pointer it is about rather than covering an `AtomicBool` that never needed it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -16,9 +16,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "1.1.4"
+version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ddd31a130427c27518df266943a5308ed92d4b226cc639f5a8f1002816174301"
+checksum = "c982642fa9e8606056828ee9a8505737230110bb1099153c79efe865c59d12ba"
 dependencies = [
  "memchr",
 ]
@@ -75,9 +75,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.102"
+version = "1.0.104"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7f202df86484c868dbad7eaa557ef785d5c66295e41b460ef922eca0723b842c"
+checksum = "330a5ed07fa54e4702c9d6c4174f74427fc0ef6e214bbd677ae50a5099946470"
 
 [[package]]
 name = "askama"
@@ -119,7 +119,7 @@ dependencies = [
  "rustc-hash",
  "serde",
  "serde_derive",
- "syn",
+ "syn 2.0.119",
 ]
 
 [[package]]
@@ -136,7 +136,7 @@ dependencies = [
  "rustc-hash",
  "serde",
  "serde_derive",
- "syn",
+ "syn 2.0.119",
 ]
 
 [[package]]
@@ -148,7 +148,7 @@ dependencies = [
  "memchr",
  "serde",
  "serde_derive",
- "winnow",
+ "winnow 0.7.15",
 ]
 
 [[package]]
@@ -160,7 +160,7 @@ dependencies = [
  "memchr",
  "serde",
  "serde_derive",
- "winnow",
+ "winnow 0.7.15",
 ]
 
 [[package]]
@@ -252,7 +252,7 @@ version = "3.4.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "290f7f2596bd5b78a9fec8088ccd89180d7f9f55b94b0576823bbbdc72ee8311"
 dependencies = [
- "event-listener 5.4.1",
+ "event-listener 5.4.2",
  "event-listener-strategy",
  "pin-project-lite",
 ]
@@ -291,13 +291,13 @@ checksum = "8b75356056920673b02621b35afd0f7dda9306d03c79a30f5c56c44cf256e3de"
 
 [[package]]
 name = "async-trait"
-version = "0.1.89"
+version = "0.1.91"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9035ad2d096bed7955a320ee7e2230574d28fd3c3a0f186cbea1ff3c7eed5dbb"
+checksum = "ae36dc4177970ef04fde5178d3e2429882def40e57a451f919c098f72baa6cec"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn",
+ "syn 3.0.3",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


jhugman:
This looks ok, on the face of it (cursory look), but could you rebase before I do a full review?

stupside:
> This looks ok, on the face of it (cursory look), but could you rebase before I do a full review?

I rebased it, but the commit is already included in the target branch, so the rebase left the PR empty. This PR has effectively been superseded and can be closed. Thanks!

Sorry for the confusion.

PR #438, opened 2026-08-09
fix(bindgen): escape */ in generated TypeScript docstrings

liamiepops:
Typescript does not allow nested block comments. Docstrings are emitted verbatim into a `/** ... */` block, so a `*/` in the content itself closes the docstring early and the rest gets parsed as code. The build fails.

This change escapes `*/` as `*\/` before wrapping. This doesn't affect target languages that do allow nested block comments.

Also adds a fixture with docstrings containing `*/` in a function, a record field, and an enum variant, covering both `format_docstring` and `format_docstring_indented`.

Tested with `cargo test -p uniffi-fixture-docstrings -- napi` on Linux. The three targets `[Jsi, Wasm, Napi]` pass.

---

**Edit:** Added unit tests to `docstring.rs` instead of adding a fixture. Tested with `cargo test -p ubrn_bindgen`.


fixtures/docstrings/src/lib.rs
```diff
@@ -0,0 +1,39 @@
+/*
```
jhugman: Might this change be better tested with a unit test instead of a fixture?

jhugman (commented):
This is great, thank you. 

A question about testing, but otherwise you'll be good to land this.

fixtures/docstrings/src/lib.rs
liamiepops: It's certainly smaller, and it won't affect coverage that I can see. I've pushed that change now, unit tests added to `docstring.rs` and removed the fixture.

jhugman (approved):
Chef's Kiss. Thank you.

PR #440, opened 2026-08-12
fix(napi): cache callback trampolines instead of rebuilding per call

jhugman:
### Reachability

**Reachable through the generated wrappers — affects every consumer of an async API.**

`async-rust-call.ts` marshals its continuation callback to `rust_future_poll` on every
poll, so ordinary `await` grew RSS without anyone touching the native module directly. No
opt-in, no unusual usage: any app awaiting a uniffi async function paid this.

(For contrast, the vtable path also builds a trampoline per call, but generated code
registers a vtable once at module init, so it is not reached through the wrappers.)

### The intended design

Callback trampolines are **deliberately never reclaimed**. The library may invoke the
function pointer from any thread at any later time, so nothing can safely free one. That
is fine, because the intended bound is **one trampoline per callback type**: a fixed
startup cost, paid once, proportional to how many callback signatures a module declares.

`callback.rs` says as much — it describes trampolines as leaked "per registration".

### The bug

Dispatch built a fresh trampoline on **every call that marshalled a Callback argument**, so
that bounded startup cost became unbounded growth.

Note the unit: the cost is per *construction* — JS→Rust, at call time — not per callback
*invocation*. One trampoline services any number of Rust→JS crossings for free. Each
construction costs:

- a strong `napi_ref` pinning the JS function,
- a leaked `CallbackUserData` (arg layout, cloned type descriptors, `Arc<Module>`),
- a fresh JS function **and** a fresh `ThreadsafeFunction` (a libuv async handle), plus an
  entry appended to the env's TSFN registry — itself unbounded,
- a leaked `TrampolineUserdata`, and
- a libffi `Closure`, which owns an executable page.

Measured in isolation, one construction costs **~2869 B** (fresh closures defeat the cache;
a cache hit measures 1 B/call and a callback-free call 0 B/call, both at n=30k).

It is reached from ordinary generated code, not just hand-written FFI:
`typescript/src/async-rust-call.ts` marshals its continuation callback to
`rust_future_poll` on **every poll**. `noopAsync` polls once per await, so there one await
is one construction — which is why the per-await figures below track the per-construction
cost.

### Measurements

Through the generated API of `fixtures/benchmark`, RSS growth per call after a forced
`global.gc()`. `noopAsync()` touches no buffers at all, so this isolates trampoline cost
from anything RustBuffer-related:

| | before | after |
|---|---|---|
| `noopAsync()` | **3455 B/await** | **7 B/await** |

Of that 3455 B, ~2869 B is the trampoline construction; the remainder is the rest of the
async protocol per await (the promise, the resolver-map entry, and the other three FFI
calls), which the cache also removes because those costs were incurred alongside it.

The shapes differ, which is the actual evidence. Before: flat at ~3.2 KB across 1k/5k/20k
iterations — unbounded. After: 139 → 121 → 7 B/await as n rises to 200k, with
non-monotonic totals (2.7 / 9.2 / 1.3 MB) — bounded V8 heap-reservation noise.

Worth being explicit, because not all RSS growth in this runtime is equal: the *per-call*
growth here was genuine and irreducible. It is `Box::leak` plus `mem::forget` — nothing
frees it, and it survives event-loop turns. (The measurement above is a loop of `await`s,
which yield on every iteration.)

### The fix

Cache one trampoline per **(callback name, JS function)**, stored on the function object
under a hidden per-callback `Symbol` — the same mechanism `CapacitySymbol` already uses to
record RustBuffer ownership on a view, so no new abstraction.

Nothing about lifetimes changes: trampolines are still never freed. There is simply one
per distinct function instead of one per call.

**Why that restores the per-type bound.** Generated code passes per-type functions, created
once:

- a callback interface's vtable is a module-level `const` whose field closures are built at
  module load, handed to `init_callback_vtable` a single time;
- `uniffiFutureContinuationCallback` is likewise a module-level `const`, reused for every
  poll of every future.

So for generated bindings, the set of distinct `(callback name, JS function)` pairs *is*
the set of callback types. Measured with the fixture:

| | distinct trampolines |
|---|---|
| one stable function, 500 calls | **1** |
| 20 freshly-created closures | **20** |

### Why keying on the function object is safe

- The Symbols belong to a single `register()` call, hence a single `napi_env`, and a JS
  function object cannot cross realms — so a cached trampoline is never handed to a
  different env than the `raw_env`/`owner_thread` its userdata captured.
- A trampoline carries no per-call state. Callbacks receive their handle as an ordinary
  argument, so reuse cannot mix up call state.

A cache hit costs only the lookup; `JsFunction::from_raw` and everything else runs only on
a miss.

### Limits

The bound is per `(type, function)`, not per type unconditionally — the second row of the
table above is the honest boundary. A hand-written caller passing a freshly created closure
on every call still gets one trampoline per call, exactly as before.

Making it per-type in all cases would mean not baking a specific JS function into the
userdata at all, and instead dispatching via the handle argument through a registry — which
is how uniffi's own model works, since each vtable method already receives a handle and the
JS side already keeps a handle map. That is a much larger change and unnecessary while
generated code passes stable per-type functions.

**Also out of scope:** vtable-struct arguments build trampolines per call too, but
generated code registers a vtable once at module init, so client code does not reach that
path.

### Note on the `Registration` struct

Adding the cache pushed `call_ffi_function` to 8 parameters, tripping
`clippy::too_many_arguments`. Since the repo has no `#[allow(clippy::…)]` anywhere, the fix
is to bundle the two per-registration Symbols into a `Registration` struct rather than add
the first suppression. That accounts for the churn in `register/mod.rs` and the signature
change — both members are keyed to the `napi_env` that `register()` ran on, so they belong
together anyway.

### Tests

Three new tests in `callback_trampoline_cache.test.mjs` assert the *invariant* — the same
function reuses one trampoline, distinct functions get distinct ones, and dispatch stays
correct across many calls — rather than sampling RSS. Per-call byte accounting is unusable
at this scale: V8 moves its heap reservation by megabytes, and a 16 KB-payload loop
measured 0.9 MB, 9.0 MB and 0.1 MB of "growth" at increasing iteration counts. Reading the
cached pointer back through the marker Symbol is white-box, but it is the only exact signal
available — the trampoline pointer is not otherwise observable from JS.

All three fail before this change; the third (correct dispatch) passes before and after, as
it should.

### Verification

macOS/arm64, Node 26: **101/101** runtime tests, **48/48** napi fixture tests,
`cargo clippy` clean, `cargo fmt --check` clean. CI green on Linux, including
`Integration tests (N-API runtime and bindings)`.

### DIFF
--- a/runtimes/napi/src/call/mod.rs
+++ b/runtimes/napi/src/call/mod.rs
@@ -51,7 +51,7 @@ pub(crate) fn call_ffi_function(
     module: &Arc<Module>,
     arg_types: &[FfiTypeDesc],
     has_rust_call_status: bool,
-    capacity_symbol: &CapacitySymbol,
+    registration: &crate::register::Registration,
 ) -> Result<JsUnknown> {
     let declared_arg_count = arg_types.len();
 
@@ -72,7 +72,7 @@ pub(crate) fn call_ffi_function(
                         env.raw(),
                         js_val,
                         module.rb_ops().from_bytes_ptr,
-                        Some(capacity_symbol),
+                        Some(&registration.capacity_symbol),
                     )?
                 };
                 slot::write_rust_buffer(slot, rust_buffer);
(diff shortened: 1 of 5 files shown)


runtimes/napi/src/call/mod.rs
```diff
@@ -86,17 +86,42 @@ pub(crate) fn call_ffi_function(
...
-                    .make_callback_trampoline(
-                        cb_name,
-                        callback::on_js_thread,
-                        callback::dispatch_to_js_thread,
-                        callback::is_js_thread,
-                        user_data,
-                    )
-                    .map_err(core_err)?;
+                let raw_fn_val = unsafe { js_val.raw() };
+                // Reuse this function's trampoline if it already has one. Building a
```
1egoman: nitpick: Maybe add a "SAFETY: " here?


```suggestion
                // SAFETY: Reuse this function's trampoline if it already has one. Building a
```

1egoman (approved):

runtimes/napi/src/call/mod.rs
jhugman: Added — thanks. I went slightly wider than the suggestion: the prose it pointed at explains the *caching strategy* rather than a safety invariant, and it sat above `raw_fn_val` while three separate `unsafe` blocks in this arm had no note of their own. So the comment block stays descriptive, and each `unsafe` now carries its own `SAFETY:`:

- `js_val.raw()` — value from the current callback scope, so `raw()` yields a `napi_value` valid for that scope without transferring ownership.
- `trampolines.get(...)` — same env, value read just above; a miss is `Ok(None)`, so an unmarked function is built below rather than misread.
- `JsFunction::from_raw(...)` — errors rather than aborting on a non-function, and the declared arg type is `Callback`, so that case surfaces as a JS exception.
- `trampolines.set(...)` — same env and value; the stored pointer is the trampoline just built for `cb_name`.

Comment-only change, but it is in `#440` so the two PRs above it were restacked. Verified each layer after the rebase: 101/106/109 runtime tests, clippy clean, and the diff against the previous heads is comments only.

PR #462, opened 2026-09-11
Solidify the runtimes/core API

jhugman:
This PR changes the API between runtimes/jsi and runtimes/core, and runtimes/napi and runtimes/core.

This is to:
- make runtimes/core and runtimes/napi more regular/uniform.
- audit the `unsafe`
- migrate regression tests introduced in #440, #441 and #442 to NAPI and JSI2.

All fixtures should still pass.

Behaviour changes, both in the direction of failing at registration rather than at the call, or of failing rather than silently degrading:

- napi `register()` now fails outright for a function whose declared return is `Struct` or `ForeignBytes`; before, only the call failed.
- A callback declared with a `Struct` return and `out_return: false` now errors in `create_callback_user_data`; before, it silently got `ret_size = 0`.
- Both bridges size a function's return buffer from core's `return_size`, which widens pointer returns to 8 bytes the way core's call path writes them. On a 32-bit host, sizing from the arg-slot geometry undersized the buffer for every pointer-returning function.
- `Module::unload` still drains and clears the trampoline map when the unloading flag is already set (by an earlier `unload` or a `disarm`); before, it returned early, so `disarm` then `unload_force` would have closed the library with the map still full.

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


runtimes/core/src/callback.rs
```diff
@@ -103,7 +103,10 @@ unsafe extern "C" fn trampoline_body(
     // If the module is shutting down, zero the return value and bail.
     if userdata.unloading_flag.load(Ordering::Acquire) {
```
zzorba, line 104: Pre-existing, but the PR now leans on this as the reload story and the new fixture asserts on the empty string it produces: a disarmed trampoline zeroes the return and leaves the `RustCallStatus` untouched, so Rust reads code 0 plus `ffi_default()` — a *successful* empty return. For a callback returning a handle, a `Result`, or anything where the default is meaningful, a silent default is worse than an error.

The RCS pointer is in `args[arg_slots.len()]` right here whenever `has_rust_call_status` is set. Writing `CALL_UNEXPECTED_ERROR` (2) through it on this path would turn the silent default into a reported failure at the Rust call site, with no ABI change.

runtimes/core/src/callback.rs
```diff
@@ -214,10 +225,80 @@ impl Module {
...
+    /// identity itself, so it cannot check this.
+    ///
+    /// Dropped once the module is unloading, which two shutdown paths lean on.
+    /// `unload` empties the map after draining, so an insert landing behind it
+    /// would hand a later lookup a pointer into a library `unload_force` has
+    /// since closed. `disarm` keeps the map instead of clearing it, and that is
+    /// only sound because this guard means nothing is added after the flag is
+    /// set: what the map holds is a fixed set of entries the flag has already
+    /// made inert.
+    pub fn remember_trampoline(&self, callback_name: &str, identity: u64, fn_ptr: *const c_void) {
```
zzorba, line 287: Small race: the flag is read before the lock is taken, so `unload` can set the flag, drain, and `clear()` the map between this check and the `insert` below, and the insert lands anyway — exactly the case the doc says the guard exists for. Moving the `is_unloading()` check under the `trampolines` lock closes it, since `unload` clears under the same lock after setting the flag.

runtimes/core/src/lifecycle.rs
```diff
@@ -70,6 +70,34 @@ impl Module {
...
+    /// One way, and it forecloses the others: `unload` keys off the same flag,
+    /// so after a `disarm` it early-returns without draining or clearing the
+    /// map, and an `unload_force` behind it would `dlclose` a library with
+    /// in-flight calls and a map still full of pointers into it. No frontend
+    /// does both today — jsi disarms and never unloads, napi unloads and never
+    /// disarms — and one that wanted both would have to give `unload` a
+    /// stronger test than the flag.
+    ///
+    /// Returns `Ok(())` immediately if the module is already unloading.
+    pub fn disarm(&self) -> Result<()> {
```
zzorba, line 93: The doc is admirably honest about the footgun, but it's a cheap one to remove: if `unload` still ran `wait_for_drain()` and cleared the map when `begin_unload()` fails (rather than early-returning), then `disarm` → `unload` → `unload_force` would be safe by construction instead of "no frontend does both today". Drain on an already-flagged module is idempotent, so there's no downside.

runtimes/core/src/module.rs
```diff
@@ -66,51 +74,85 @@ impl ResolvedFunction {
...
+                let rb: RustBufferC = unsafe { self.cif.call(code_ptr, ffi_args) };
+                put_return_bytes(out, &crate::slot::rust_buffer_to_bytes(&rb))
+            }
+            FfiTypeDesc::VoidPointer
+            | FfiTypeDesc::Reference(_)
+            | FfiTypeDesc::MutReference(_)
+            | FfiTypeDesc::Callback(_) => {
+                // SAFETY: as in `call_and_put!`, with usize for the CIF's pointer return type.
+                let v: usize = unsafe { self.cif.call(code_ptr, ffi_args) };
+                put_return_bytes(out, &(v as u64).to_ne_bytes())
```
zzorba, line 136: Pointer returns are widened to 8 bytes here, but both bridges size their return buffer from `slot_size_align`, which reports *pointer* width for `VoidPointer`/`Reference`/`MutReference`/`Callback` (napi: `register/mod.rs` L94; jsi: `ubrn_jsi_scalar_slot_size_align`). On the 32-bit Android ABIs the rest of this PR takes care to support, a pointer-returning function would fail `put_return_bytes` with "return buffer too small". The jsi shim mostly dodges it by rewriting `VoidPointer` → `Handle`.

Suggest a `return_width(desc)` (or similar) in core that both bridges size from, so the width the writer uses and the width the reader allocates come from one place.

runtimes/core/src/module.rs
```diff
@@ -66,51 +74,85 @@ impl ResolvedFunction {
...
         }
+        ffi_args
+    }
+
+    /// Call the resolved symbol, writing the return value's native-endian bytes
+    /// into `out`. Returns the number of bytes written (0 for a void return).
+    ///
+    /// A pointer return is widened to 8 bytes so the byte width does not vary
+    /// with the host's pointer size. A `RustBuffer` return is written as its
+    /// 24-byte `repr(C)` form and stays owned by the caller, who must free it —
```
zzorba, line 85: Nit: the code writes `size_of::<RustBufferC>()`, which is 20 on i386 (the `abi_assert.cpp` comments spell this out), not always 24. The shim's `uint8_t out[sizeof(UbrnRustBuffer)]` is correct; just the doc here (and the `ubrn_jsi_call` doc) is 64-bit-specific.

cpp/jsi-player-shim/callbacks.cpp
```diff
@@ -498,71 +571,184 @@ extern "C" void cb_on_js_thread(const uint8_t *args, uint8_t *ret,
...
+  slot->ret.resize(shape.retSize);
+
+  const void *capturedUd = udPtr;
+  auto captured = state;
+
+  // Rendezvous: post onto the JS thread via invokeAsync, block until it runs or
+  // the module is aborted. NEVER invokeSync (it is a no-op in the test
+  // harness). uniffi foreign-thread callbacks fire from Rust-owned threads, so
+  // the JS thread is free to drain.
+  ud->callInvoker->invokeAsync(
```
zzorba, line 687: The posted task never re-checks `aborted`. A worker released by `abortModule` returns to Rust and pops the frame that the `out_return`/`RustCallStatus` pointers inside `slot->args` point at, while this queued lambda still holds those pointers plus the raw `capturedUd`. The comment below says no queued task ever runs because the host discards the queue first — that's true of `test-runner.cpp`, but it's a host precondition the shim doesn't establish, and it doesn't hold if the root is finalized mid-runtime.

Cheap fix that makes the claim true by construction: at the top of the lambda, lock `captured->mtx`; if `captured->aborted`, set `slot->done`, `notify_all`, and return without calling `on_js`.

cpp/jsi-player-shim/callbacks.cpp
```diff
@@ -498,71 +571,184 @@ extern "C" void cb_on_js_thread(const uint8_t *args, uint8_t *ret,
...
 }
 
+// Core invokes this through a plain `extern "C"` fn pointer, which rustc marks
+// nounwind — on the same-thread path it is called straight from
+// `trampoline_body`. A C++ exception crossing that frame is undefined
+// behaviour, so nothing may escape here. Every synchronous vtable method is
+// out_return, so ret_size is 0 here: Rust reads back FfiDefault::ffi_default()
+// with call_status.code still 0, a successful empty return, not a reported
+// failure.
+extern "C" void cb_on_js_thread(const uint8_t *args, uint8_t *ret,
```
zzorba, line 581: Turning an unwind through a `nounwind` frame into `catch(...) {}` is right, but the outcome is now a silent *success*: `statusPtr->code` stays 0 and a non-`out_return` return stays zero-filled. "Every synchronous vtable method is out_return" describes the common case rather than an invariant the code enforces (the non-out_return + hasRcs path exists a few lines up). What actually reaches this barrier is shim-side marshalling failure (`unknown struct`, consumed `RustBuffer`, bad pointee) — the cases where reporting success is worst.

Suggest resolving the RCS pointer before the `try` and writing `CALL_UNEXPECTED_ERROR` (2) on catch when one is available; at minimum, log.

cpp/jsi-player-shim/value_conv.h
```diff
@@ -15,174 +15,150 @@
...
+  if (tagName == "MutReference")
+    tagName = "Reference";
+  d.tag = ubrnTagFromName(tagName);
   // A VoidPointer is pointer-sized; we treat it like a Handle for byte layout.
-  // (Appears as a callback's out_return arg type / void return; never needs a
-  // name.)
-  if (tag == "VoidPointer") {
+  // A shim-local parsing choice, not an ABI tag. (Appears as a callback's
+  // out_return arg type / void return; never needs a name.)
+  if (d.tag == UBRN_TY_UNSUPPORTED && tagName == "VoidPointer") {
```
zzorba, line 118: Renaming `VoidPointer` → `Handle` changes the CIF type core builds from *pointer* to *u64*. Identical on 64-bit; on 32-bit a `void*` arg or struct field described as u64 gives a wrong CIF (4- vs 8-byte arg, shifted struct offsets). Probably dormant now that `ret_tag_for_out_return` emits `Void` and core appends the out-return slot itself, but core accepting `VoidPointer` (or a `Reference`-style pointer rewrite) would be safer than a rename that changes the width.

runtimes/napi/src/call/mod.rs
```diff
@@ -134,13 +149,26 @@ pub(crate) fn call_ffi_function(
...
-                            registration
-                                .trampolines
-                                .set(env.raw(), raw_fn_val, cb_name, fn_ptr)?
+                        // Minted and stashed only here, past the function check: `set`
+                        // writes a hidden marker onto the caller's own value, and a value
+                        // that turns out not to be a function is handed back untouched.
+                        let identity = match stashed {
+                            Some(identity) => identity,
+                            None => {
+                                let identity = registration.trampolines.next_identity();
```
zzorba, line 158: Order here is build trampoline → mint identity → `set(...)?` → remember. If `set` fails (frozen/non-extensible function, as `cache.rs` notes), the just-built trampoline, its leaked userdata, the strong `fn_ref` (which pins the JS function forever) and its TSFN are all leaked and never remembered; a retry leaks again. Stashing the identity *before* building is strictly safer: a later build failure then just leaves a stashed identity with no map entry, which the miss path above already handles.

runtimes/napi/src/register/mod.rs
```diff
@@ -84,6 +87,11 @@ pub fn register(
         })?;
         let arg_types: Rc<Vec<FfiTypeDesc>> = Rc::new(func_def.args.clone());
         let has_rust_call_status = func_def.has_rust_call_status;
+        // Resolved here rather than per call: the return width never changes,
+        // and looking it up in the closure would re-hash the function name on
+        // every invocation.
+        let ret_desc: Rc<FfiTypeDesc> = Rc::new(func_def.ret.clone());
+        let (ret_size, _) = uniffi_runtime_core::slot_size_align(&ret_desc).map_err(core_err)?;
```
zzorba, line 94: See the comment on core's `invoke`: for pointer-typed returns this yields pointer width, but core writes 8 bytes, so on a 32-bit host every pointer-returning function would fail with "return buffer too small". Harmless on today's Node targets; flagging because the PR otherwise takes 32-bit seriously.

typescript/src/handle-map.ts
```diff
@@ -14,6 +14,10 @@ export class UniffiHandleMap<T> {
   // "Foreign handles are generated with a handle map that only generates odd values."
   // "Foreign handles must always have the lowest bit set"
   // "0 is an invalid value."
+  //
+  // Numbering restarts here in every runtime, so handles collide across a
```
zzorba, line 18: Recording this is fair for the PR's scope, but it's a memory-safety issue on reload rather than a curiosity — the fixture has to `mem::forget` the previous stash to avoid it. A small follow-up: seed `currentHandle` from a process-global epoch the player exposes (a counter that survives the runtime), so handles from a dead runtime can never collide with live ones.

zzorba (commented):
**Provenance note:** this review was produced by an AI code review (Claude, run by @zzorba's account via Claude Code) that read the full diff, checked the flagged claims against the head commit, and looked at CI. Treat it as a first pass from a careful reader, not as a human sign-off. Nothing here was run locally: the review is by reading, and CI on the head commit is green.

**Overall:** solid PR. Pulling the tag vocabulary, callback layout, trampoline map and return path into core removes real duplication; the unsafe audit is honest (every SAFETY comment checked is backed by the code); and the reload-safety fixture covers a hazard that previously had no test. The inline comments are the actionable items. A few things that didn't fit on a diff line:

- **`RustCallStatus` struct-field arm hard-codes 64-bit offsets** (`cpp/jsi-player-shim/callbacks.cpp` ~L236, unchanged by this PR): `memcpy(slot + 8, ...)` guarded by `slotSize >= 32`. On i386 the real layout is `code@0`, buffer at 4, 24 bytes total, so the guard fails and the error buffer is silently dropped (code still lands, so Rust sees an error with an empty message). After the `abi_assert.cpp` rewrite this arm is the odd one out; `offsetof`/`sizeof` on the existing `ubrn_cb::RustCallStatus` mirror would fix it.
- **`runtimes/napi/tests/callback_trampoline_cache.test.mjs` is now vacuous for the reuse property.** It asserts the hidden marker is stable across calls, but the marker is now the *identity*, which is stable by construction even if core's map misses and a trampoline is built every call. `$uniffiTrampolineCount` is the right assertion; only the fixtures test uses it today.
- **Behaviour changes not in the description** (both arguably fixes, worth a line): napi `register()` now fails outright for a function whose declared return is `Struct`/`ForeignBytes`, where before only the call failed; and a callback declared with a `Struct` return and `out_return: false` now errors in `create_callback_user_data` instead of silently getting `ret_size = 0`.
- **`buildShape` treats every `-1` from `ubrn_jsi_callback_arg_layout` as the benign by-value-struct case**, including unknown-callback. Unreachable today, but a typo between the DEFINITIONS callbacks table and a vtable field would surface as a silently no-op trampoline rather than a registration error.

Not found: cross-runtime trampoline reuse (each `register` makes a fresh `Module` and map, identities are process-global and monotonic), any SAFETY comment contradicted by its code, or a regression in return-byte decoding versus the removed `CallReturn` code in either bridge.

fixtures/reload-safety/tests/bindings/test_reload_parked_worker.ts
```diff
@@ -0,0 +1,87 @@
...
+  type Greeter,
+  noteIteration,
+  pingerReturned,
+  pingerStarted,
+  startForeignThreadPinger,
+} from "@/generated/reload_safety";
+
+theModule.initialize();
+
+const BUSY_MS = 3;
```
zzorba, line 32: This test is timing-based (3 ms busy loop, 200 ms warm-up, bounded 2 s wait for the worker to record its exit). It's bounded rather than racy, so it should be fine, but worth watching for flakes on slow CI runners — the `pings > 5` precondition in particular depends on the JS thread getting scheduled promptly.

zzorba (approved):
LGTM, but my bot found a number of possible follow up items (not sure the quality of them TBH)

typescript/src/handle-map.ts
jhugman: Filed #470.

jhugman:
Thanks @zzorba . 

Your bot made some good points; I marked the ones I've addressed with a 👍 .

The remaining ones weren't user-accessible (neither the `VoidPointer` rename or the logging is triggerable via user code), so I'm not sure it's worth an issue yet.

PR #463, opened 2026-09-11
feat(jsi2): @ubjs/react-native, the JSI player as an npm package

jhugman:
The JSI player (`Jsi2` flavour) as an npm package, `@ubjs/react-native`. An app installs it once as a direct dependency; its C++ shim compiles inside the app's own iOS and Android builds against the app's JSI headers, and its Rust half (dispatch, libffi, `dlopen`) ships prebuilt as a shared library per slice: a `UniffiRuntimeJsi.framework` inside a dynamic xcframework that CocoaPods embeds and signs, and `libuniffi_runtime_jsi.so` per ABI packaged as `jniLibs`. About half a megabyte per slice; no Rust toolchain for app developers.

**Commits**

1. **Resolve the library by name.** Generated bindings pass `{ name }` to `globalThis.uniffi.open` and the host resolves it. The name is the built library's cdylib name, shared by every module generated from it (naming each uniffi namespace's crate opened one cdylib per namespace and aborted on the ext-types fixture). `registerNatives(rt, invoker)` keeps its two-argument signature for v1 and the Hermes test-runner; `ubrnRegisterPlayer` carries the resolver. The shim moves to `runtimes/jsi/cpp`.
2. **The package.** `package.json`, the TurboModule spec and entry that installs on import, the ObjC++ and Kotlin installers with their resolvers, podspec, `Package.swift`, and a pack test that pins what the tarball ships.
3. **The prebuilt Rust half.** Build and assemble scripts, a completeness test, and the publish workflow. Shared rather than static: a static archive carries the whole std closure the consumer's linker discards (18 MB vs 0.45 MB). The Android `.so` carries a soname, verified at build time, so the shim's `DT_NEEDED` is a bare name.
4. **Compile gate.** A fresh React Native app with the packed player as a direct dependency builds for the iOS simulator and Android on CI, asserting the framework is embedded and both `.so` files are packaged.
5. Changelog.

6,400 lines of code is the package-lock.json file.

**Not in this PR:** generating libraries that ship through the player (`ubrn build jsi2`, `ubrn generate jsi2 all`) is the next PR in the stack. Nothing has yet been launched on a simulator or emulator; that lane is the first task of the next PR. The `lift` parity precondition (#458) is still open; the reload-signpost cleanup it enables is a tracked follow-up.

**Known deferred minors** (from review, all non-blocking): the `ci.yml` fmt gate does not cover C++ and the repo has no `.clang-format`; the podspec's `c++17` key was dropped because `install_modules_dependencies` overwrites it; the xcframework carries no headers, which only SwiftPM consumers would notice.


runtimes/jsi/README.md
```diff
@@ -0,0 +1,37 @@
+# @ubjs/react-native
+
```
zzorba, line 2: When we updated the getting-started guide, we should make sure to include adding this dependency if you are targeting JSI.

zzorba (approved):
LGTM

PR #464, opened 2026-09-11
feat(jsi2): ubrn build jsi2 and generate jsi2 all, the library side of the player

jhugman:
The library side of the JSI player. A library author runs `ubrn build jsi2 android|ios --and-generate` and gets an assets-only React Native library that loads through `@ubjs/react-native`: TypeScript, a `lib<name>.so` per Android ABI under `jniLibs`, and a dynamic `<name>.xcframework` of `<name>.framework` bundles that CocoaPods embeds and signs. No per-library C++, Kotlin, CMake or codegen spec.

**What is here**

- `scripts/lib/rn-app.sh` + `scripts/test-player-package.sh --run`: a shared harness that scaffolds a plain RN app, launches it on the iOS simulator or an Android emulator with Metro serving, and waits for a console sentinel. The player alone now proves `globalThis.uniffi` exists in a running app on both platforms; CI runs the iOS half.
- `jsi2:` in `ubrn.config.yaml` (`ts`, `androidTargets` default `arm64-v8a`/`x86_64`, `minIosVersion` default `15.1`, `bundleIdPrefix`), and `ubrn_common::so_extension` learns iOS shared builds are dylibs.
- `ubrn generate jsi2 all`, through the existing generate-all path with a new `Platform::Jsi2`: bindings that open the library by its cdylib name, `src/index.tsx` (imports the player first), a podspec vendoring the xcframework and depending on `UbjsReactNative`, a `build.gradle` packaging `jniLibs` and depending on `:ubjs_react-native`, an empty Java `ReactPackage` (Android autolinking drops a library without one), and an order-preserving `package.json` patch adding the player as peer and dev dependency and `@ubjs/core` as a dependency.
- `ubrn build jsi2 android` (`cargo ndk`, 16 KB pages, soname) and `ubrn build jsi2 ios` (dylib per target with `IPHONEOS_DEPLOYMENT_TARGET`, lipo'd simulator slice, framework bundles with install name and `Info.plist` incl. `CFBundleSupportedPlatforms`, `xcodebuild -create-xcframework -framework`), nested under `build` like every other flavour.
- `scripts/test-jsi2-library.sh`: scaffolds a library around `examples/arithmetic` with create-react-native-library, builds and generates it, packs it, installs it and the player into a fresh app, and proves `add(2, 3)` = 5 on the iOS simulator (CI, `jsi2-library.yml`) and an Android emulator (local, against the developer's AVD).
- `docs/src/reference/jsi2.md` with the manual device release check (`codesign -dv`).

**Found along the way, fixed one PR down (#459):** the player's Android slice built on a macOS host was not loadable: libffi's configure picked Apple's `ar` and produced an empty archive, and compiler-rt's `__clear_cache` was never linked. The compile gate could not see it; the emulator lane did.

**Not in this PR:** the runtime guard in the generated entrypoint cannot fire in the case its message names, because the player's `getEnforcing` throws first; the spec's § P1.5 guard 3 should move that message into the player. P3 (single-compile mode) is designed for, not built.


zzorba (approved):
LGTM

PR #468, opened 2026-09-16
chore(uniffi) bump uniffi to 0.32

Dzejkop:
Hey! I'm blocked by the uniffi mismatch at the moment so decided to put a bunch of agents to work on fleshing out https://github.com/jhugman/uniffi-bindgen-react-native/pull/409 (thanks @vivianjeng!).

I ran tests & checked it against my browser wasm32 setup and everything seems to work.

I did some review of my own but I don't really have context. 

I also snuck in a flake.nix + flake.lock - for my own testing comfort, lmk if you're fine with that.


zeenix:
@Dzejkop Thanks for doing this, I was about to prepare a PR for this as I also need it in a project. BTW, it's best to rebase on main branch so that the PR only contains/shows changes from the PR, making it easier for maintainers to review. Just a friendly suggestion (I'm not the maintainer here). :)

Dzejkop:
> @Dzejkop Thanks for doing this, I was about to prepare a PR for this as I also need it in a project. BTW, it's best to rebase on main branch so that the PR only contains/shows changes from the PR, making it easier for maintainers to review. Just a friendly suggestion (I'm not the maintainer here). :)

Thanks! I didn't want to lose commits from @vivianjeng since they started the work. Up to the maintainers imo - I can squash if necessary

zeenix:
> Thanks! I didn't want to lose commits from @vivianjeng since they started the work

Understandable but you don't loose commits or their authorship from a rebase, unless you choose to.

neekolas:
I blew a bunch of tokens reviewing this PR with Opus 5.5 and running all the tests. Results below: 

## Summary

The core of the upgrade is correct. The 0.32 IR changes, the `GlobalConfig` and cargo_metadata changes, reading explicit discriminants from the `initial` IR, and the `ForeignBytes` plumbing work on all four flavours. Borrowed `&[u8]` args run correctly on a real iOS build. But there are **3 regressions against `main`**, a **stale `Cargo.lock`**, and **the RN compat CI job will fail**. There is also a lot of scope beyond the upgrade (Set/Box/recursive enums, `rename`/`exclude`, a changed tagged-enum type shape, a Nix flake), and that makes the change hard to review. I suggest we request changes.

## What I ran

All runs were local (macOS arm64, rustc 1.98.1, wasm-bindgen 0.2.114, Hermes rn/0.77-stable), on this branch at `06f0a30` and on `main` at `f5865e7`:

| CI step | `main` | this PR |
|---|---|---|
| fmt + clippy, `cargo build`, `tsc` | ✅ | ✅ |
| Unit tests (workspace, CI excludes) | 98 pass | 128 pass |
| `jsi::` (Hermes) | 41 pass | 43 pass |
| `napi::` / napi runtime `npm test` | 48 / 109 pass | 50 / 116 pass |
| `wasm::` | 40 pass | 41 pass |
| `wasm2::` / wasm runtime `npm test` / `wasm2_build` | 24 / 42 / 1 pass | 25 / 55 / 1 pass |
| RN compat: `test-turbo-modules.sh --ios --pack` (RN 0.87.1, bob 0.63.1) | ✅ | ❌ (see 4) |

- The per-flavour test lists are a strict superset of `main`: +array_buffers on all four flavours, and +borrowed_bytes_no_copy on jsi and napi. `Cargo.lock` has only 0.32.1 uniffi crates.
- I moved the compat starter crate to 0.32.1 locally and added `&[u8]` functions and methods. The turbo module then built with Xcode and ran in an iPhone 16 simulator. 9/9 checks passed, including an empty buffer, a 100 kB buffer and 1000 calls in a loop.

## Blocking issues

**1. Proc-macro-renamed objects break on JSI.** `api_module/builders.rs:753-761` and `ffi_module/builder.rs:76-81` build the `bless_pointer` symbol from `interface.orig_name`. In 0.32, `orig_name` is the Rust identifier, but the C++ side uses the foreign name from the ComponentInterface.
- Repro: add `#[derive(uniffi::Object)] #[uniffi(name = "RenamedWidget")] struct RustWidget` to enum-types. The C++ registers `ubrn_uniffi_internal_fn_method_renamedwidget_ffi__bless_pointer`, but the TS calls `…_rustwidget_ffi__bless_pointer`, which does not exist at runtime. On `main`, every file uses `renamedwidget`.
- The existing test covers only a uniffi.toml rename.
- Fix: use the same source the C++ side uses (for example, derive from `ffi_func_free`), and add a fixture for `#[uniffi(name = …)]`.

**2. A flat `--config` now replaces each dependency crate's config in the TS output.** `cli.rs:236-266`: `ConfigOverrideLayer` returns the same file for every crate, and the first layer wins. The TS pipeline now uses this loader. On `main` it deliberately did not, to prevent exactly this. `ubrn` passes `--config` whenever `ubrn.config.yaml` sets `bindings.uniffiToml`, so real projects are affected.
- Repro: generate ext-types with a flat `--config` that contains only `strictTypeChecking`. `custom_types.ts` changes from `export type Url = URL` to `export type Url = string`, and the custom converter import is dropped.
- Fix: keep `main`'s rule that the TS pipeline uses per-crate configs, or apply the override only to the root crate. Add an ext-types test with `--config`.

**3. `Cargo.lock` is stale.** It looks like the lock from #409, merged over `main`'s lock. It downgrades about 80 unrelated crates (serde 1.0.229→1.0.228, tokio 1.53.1→1.50.0, clap 4.6.5→4.6.0, cc, getrandom, icu_*, and more) and adds 11 crates that `main` does not have (wit-bindgen*, wasip2/3, …). Starting from `main`'s lock and applying this PR's `Cargo.toml` files changes only 17 packages. Fix: `git checkout main -- Cargo.lock`, then `cargo update -p` for the uniffi crates and cargo_metadata.

**4. The RN compat CI job will fail.** `integration/fixtures/compat/ubrn.config.yaml` still points at `uniffi-starter` branch `jhugman/bump-uniffi-to-0.31`. The 0.32 bindgen cannot read 0.31 metadata:
```
extracting metadata for '_UNIFFI_META_FOOBAR_CONSTRUCTOR_CALCULATOR_NEW'
Unexpected metadata type code: 116
```
A 0.32 branch of uniffi-starter is needed, with this fixture pointing at it.

**5. Rejecting `&[u8]` together with a callback-interface argument is a regression.** `api_module/mod.rs:672-700` rejects `fn f(b: &[u8], cb: Box<dyn Cb>)`. uniffi 0.32.1 compiles this (I checked with a probe crate), and `main` supports it.
- The guard also does not give the safety its error message claims. It checks only arguments, so `fn feed(&self, b: &[u8]) { self.cb.call() }` passes the check and still re-enters JS.
- It also rejects on wasm/wasm2, where the bytes are already a copy.
- Please drop the `takes_callback` branch.
- The callback-method branch cannot be reached, because uniffi rejects `&[u8]` in callback/foreign trait methods with E0053. The same applies to `CppFfiCallbackFunctionExt::has_foreign_bytes_args`, which no template uses.

## Keep, but reword

**The async `&[u8]` rejection is correct, but for a reason the comment does not give.** On native targets, uniffi 0.32.1 already refuses `async fn f(b: &[u8])` (`*const u8` cannot be sent between threads). With `uniffi/wasm-unstable-single-threaded` on `wasm32-unknown-unknown`, it **compiles**. The future then reads the pointer after the sync call returns, and wasm2 has already freed the copy by then. The error message should say this.

## Non-blocking

- **The tagged-enum TS type changes for every enum, not only recursive ones** (`TaggedEnumTemplate.ts:14-20`). `export type X` changes from `InstanceType<…>` to a union of `X_V_interface` literals, `tag`/`inner` lose `readonly`, and the enum's docstring moves to the first private alias. Suggestion: apply it only when the 0.32 `recursive` flag is set, keep `readonly`, and keep the docstring on `export type X`. Mention it in the CHANGELOG.
- **`BoxRenameFix` (`cli.rs:287-440`) works around an upstream bug:** `general::map_type` does not rename inside `Type::Box`. That is fine for now, but please file an upstream issue and add a link in a TODO. It is needed only for `rename`, which is new in this PR.
- **SharedArrayBuffer:** a SAB-backed view now throws on JSI (`ForeignBytes.h:81`) and napi (`napi_utils.rs:398`). `main` accepted it by copying, and wasm/wasm2 still accept it. Copying SAB input into a temporary keeps the behaviour the same.
- **wasm2 borrow-in-place** (`call.ts:196-210`): the `v.buffer === memory.buffer` path runs only for user-made views over wasm memory. On pre-ES2024 engines, a view that an earlier argument's `memory.grow` detached becomes an empty slice without an error. Suggestion: always copy.
- **Detached-buffer behaviour differs by flavour.** napi throws (which also forced the napi7 bump). JSI and older wasm2 engines pass an empty slice.
- **Consumed-view markers** (JSI `ForeignBytes.h:76`, napi) guard views that user code never gets, because lifts copy. About 8 JSI calls per argument. A strong `jsi::ArrayBuffer` plus a bounds check is enough.
- **Leaks on the new throw paths:** owned RustBuffer args that were converted before a `ForeignBytes` validation throws are not freed (JSI `macros.cpp` phase 1, napi first loop).
- **Generated `-ffi.ts`:** the comment "We perform a compile time check…" and the `RuntimeUniffiRustFutureContinuationCallback` import remain after the check they served was removed.

## Scope

These are not part of the 0.32 upgrade. Please split them into their own PRs or drop them:
- `flake.nix` / `flake.lock`
- the `"./package.json"` export in `typescript/package.json`
- the `0.31.0-3` → `0.31.0-5` doc version bumps
- `Set` / `Box` / recursive-enum support, and `rename` / `exclude` (with `BoxRenameFix`)

Set and Box were not usable in 0.31, so for the upgrade itself a clear "not yet supported" error is enough, and no feature is lost. The Set wire format here does match `uniffi_core` (i32 count, then items), so that code is a good base for a follow-up.

The CHANGELOG cites #409 rather than this PR, and does not mention Set/Box or the enum type change. A squash/rebase onto `main` would also make the history much easier to review. Right now it has 30 commits, including 3 merge commits from #409.

jhugman:
Sorry, I've been somewhat absent in this discussion. Thanks for the work @Dzejkop @zeenix @neekolas .

I currently working on #360 and #412 which I want to get out for a 0.31.x release. I'm super happy to support you doing a 0.32.x upgrade, though I think we should get you on a release branch, then forward port the version bump on top of main on top of the #412 work, rather than backporting #412 on to `0.31.x`. What say you, is this okay?

